PR #193, opened 2014-08-12
Handle ldap search

sabade:
Many LDAP implementations require looking up values in order to properly form the authentication bind.  This change allows for specifying a search string against the LDAP and a search used to pull the proper DN for a given user in these environments.  It also updates the command line management tooling to allow for specifying this new option via the manage.py command.


frbapolkosnik:
@sabade, can you check if the recently merged LDAP change works for your usecase?
Thanks!

PR #228, opened 2014-09-02
Disassembly

wxsBSD:
We have been kicking around the idea of storing disassembly output in CRITs, and this is a first crack at it.

Here's the basic idea:

1. Store disassembly files as a TLO and the backing binary in gridfs.
2. Keep it generic so we aren't tied to only one kind of disassembler.
2. Store different "versions" of a disassembly, similar to raw data.
3. Any analysis of the disassembly file is to be done in a service.

By storing different versions of a disassembly we can allow people to collaborate easier on RE tasking. It is a nice thing to be able to take someone's RE output and enhance it further, and then upload as a new version. If it turns out your work was completely wrong you can delete the version or tell others to use the previous version and not yours.

Keeping the analysis of the disassembly files as various services (probably distributed only so we don't have to require various licensed software running on your webserver) allows us to experiment more with difficult tasks.


wxsBSD:
I'm going to close this as I don't think it's ready yet and I don't have the time to keep it up right now. Will reinvestigate if time permits in the future. If someone wants to either continue this branch or do something entirely different please feel free.

PR #313, opened 2014-11-24
LDAP bind_dn support

frbapolkosnik:
Support for two-step LDAP binding using bind_dn and bind password for the first bind and search of the user's proper user_dn based on the cn or uid supplied by the user at the login prompt.

In many places each webapp using LDAP receives its own dedicated bind_dn and a password, so that lookups can be performed for cn or uid of user accounts (and anonymous binds are not required to be used), from there a proper user_dn is obtained, and proper authentication can occur for users across different LDAP servers.

This change does not impact the current LDAP setups that don't use the 2-step process, yet it fixes an issue that I had for quite a while.

I welcome any questions or suggestions.


mgoffin:
Looking at the diff I'm not quite sure what happened but it looks like it's trying to apply fixes that were in other commits? Specifically the Campaign changes in crits_mongoengine and the S3 changes.

Also is there any way to de-duplicate this code so the same ldap binding process isn't in two places?

frbapolkosnik:
The reason for two places is that one is for plain LDAP checkbox case, the other one is for user details update. I'm not that proficient with git yet, and the merge to fix the conflicts when the whitespace fixes went in caused this.

frbapolkosnik:
I can definitely re-do it in another PR to clean it up.

mgoffin:
I know one is for one use case and one is for another, but if they are using the same exact code, that smells like an opportunity to write a function and have them both call it? :)

frbapolkosnik:
I'll try to find some time to collapse it. To make things cleaner I'll close this PR, and  create a separate PR for the updated version.

frbapolkosnik:
I think that I've fixed the miss-merge, next stop will be to de-duplicate the ldap code.

frbapolkosnik:
Just an FYI, this is still on my TODO list.

frbapolkosnik:
Might as well merge this.

PR #333, opened 2014-12-03
API delete support for domains and IPs

puhley:
This is a rough draft of code that will enable delete support for the IP and domain APIs. Since an IP or domain can be associated with multiple sources and campaigns, this approach allows you to delete an individual reference or to delete the entire record. If you provide a campaign and source along with the record ID, then this will just delete the references to that campaign and source. If you only provide the record ID, then it will delete the entire record.

The code for delete support in the domain API is a little complicated. The django-tastypie-mongoengine won't accept a DELETE request with parameters passed in the body of the request. Therefore, the DELETE request must pass the parameters in the URL. However, in order to pass the request to the generate_domain_jtable, the request must be converted into a POST format. Since I didn't know the jtable code well enough to reformat it, I got around the issue by creating a fake POST object and passing that instead. This isn't the most elegant solution but it allowed me to contain the changes to the API code.

Being able to delete an IP or API reference would be useful for people who want to use CRITs to track only the current list of threats. For instance, IPs are sometimes removed from IOC lists because the compromised host has been fixed. These changes would allow you to remove the IP from the current list of threats via the API. This code still requires the user to have admin privileges in order to perform the deletion.

Please take a look at the code and let me know your thoughts.

### DIFF
--- a/crits/campaigns/api.py
+++ b/crits/campaigns/api.py
@@ -12,8 +12,6 @@ from crits.indicators.indicator import Indicator
 from crits.ips.ip import IP
 from crits.pcaps.pcap import PCAP
 from crits.samples.sample import Sample
-from crits.raw_data.raw_data import RawData
-from crits.actors.actor import Actor
 
 from crits.campaigns.handlers import add_campaign, campaign_remove, remove_campaign
 from crits.core.api import CRITsApiKeyAuthentication, CRITsSessionAuthentication
@@ -76,12 +74,12 @@ class CampaignResource(CRITsAPIResource):
             content['message'] = 'Need a Campaign name.'
             self.crits_response(content)
 
-        result = add_campaign(name,
-                              description,
-                              aliases,
-                              analyst,
-                              bucket_list,
-                              ticket)
+        result =  add_campaign(name,
+                               description,
+                               aliases,
+                               analyst,
+                               bucket_list,
+                               ticket)
         if result.get('id'):
             url = reverse('api_dispatch_detail',
                           kwargs={'resource_name': 'campaigns',
@@ -186,7 +184,6 @@ class CampaignResource(CRITsAPIResource):
         """
         This will delete a specific campaign ID record.
         It will also delete all the campaign references in the TLOs.
-        This will override the delete_detail in the core API.
 
         The campaign ID must be part of the URL (/api/v1/campaigns/{id}/)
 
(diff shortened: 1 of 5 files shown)


puhley:
* I changed the logic from “removing the campaign from the domain” to “removing a domain from the campaign”. This eliminates the redundant campaign removal logic for each TLO. The new code will allow you to remove any type of object from a campaign.
* I added the ability to delete a campaign. I saw that functionality requested in one of the threads.
* All of the TLOs used jtable_ajax_delete. Therefore, I split jtable_ajax_delete into 2 functions so that both the UI (via jtable_ajax_delete -> delete_id) and the API (via a direct call to delete_id) could reference the same deletion code for ObjectIds.
* I added validate_objectid to the mongo_tools library to centralize validation code for ObjectIds.

dsnellgrove:
is the validate_objectid redundant with the bson library's built-in validator? http://api.mongodb.org/python/current/api/bson/objectid.html#bson.objectid.ObjectId.is_valid

puhley:
Thanks for the reference. I replaced validate_objectid with the bson validator.

puhley:
To make things easier, I have temporarily posted an alpha version of the scripts that I am using to test these APIs here: https://github.com/puhley/crits_api_scripts. These are the scripts which will keep CRITs in sync with open-source IP lists. There are also individual command lines for deleting information via the APIs. All of the code for making REST requests to these APIs is in libs2/crits.py.  This should allow you to see what the corresponding client side code might look like and to test the server code. Let me know if you have any questions.

mgoffin:
Nice! I'll give these a look over as well.

puhley:
Of course, right after I sent the link to my project I realized I missed a bug fix. If you have already cloned, then please re-clone.

puhley:
If it helps, I can put together documentation for these features. However, I am not sure what is the best way to provide the documentation to the project. One option is that I can paste it in here as a comment. The team would be able to copy the comment into the official documentation whenever this feature progresses to an official release. Would that work?

Also, let me know if you there are any further code improvements that you would like to see.

mgoffin:
The comment I made on Dec 22 doesn't seem to have been addressed. The problem with the way this is done is that there's tons of code duplication, and the code isn't centralized where it can be used by other TLOs which would also need to use it. It also limits patching to specific campaign/source removal.

After discussing "updating" via the API, I think #383 is the better way to go. And for removal of TLOs, the code needs to be centralized better.

puhley:
I thought I had addressed the code duplication by refactoring how campaigns are handled, switching the jtable code so that it could be used by both the API and the GUI, etc.  Most of the other code that is duplicated is just data validation on the supplied parameters. Did you want the data validation removed?

puhley:
If it makes things easier, I can close this pull request since we have #383 and #430 open. Should I leave it open for historical purposes or should I go ahead and close it?

PR #336, opened 2014-12-04
user_agent config setting for outbound communications

frbapolkosnik:
Just an additional convenience setting, so that you can slap on a custom UAS on your CRITs outbound connections (the next step will be to update crits_services to use this setting)


mgoffin:
I don't think I'm a fan of this.

apolkosnik:
Piping via a UAS changing proxy seems like a better idea.

PR #346, opened 2014-12-09
Default method/reference parameters to empty string

brlogan:
Some functions define the default of their Source Method and Reference fields to None instead of empty string (''). This sometimes leads to those fields being populated in CRITs with the actual text "None". However, if the data is pulled from an upload form, the field is populated with an empty string. It would be better, and more consistent, for these fields to simply be left blank if no data is provided, so we don't end up with some blank fields and some None fields.

Anybody see a problem with this?


brlogan:
Any additional thoughts on this? We've been using this change in Production for awhile now and have been happy with it.

mgoffin:
Can you give this a merge and I'll give it another test?

brlogan:
Okay! Freshly merged.
I had to disable the CSRF middleware to test because I was getting 403s every time I tried a POST.

mgoffin:
Evidently it needs another merge after accepting your other stuff, haha.

brlogan:
Haha, of course it does. Give it another try, please.

PR #368, opened 2015-01-14
Better debugging when debug mode is not turned on.

mgoffin:
This change exposes the Exception name and the Value in the custom
500.html template. Created an "errors.py" file which contains the
custom_500 function to facilitate populating the template. This function
also formats the traceback and write it to the "crits.log" file as
DEBUG. This will allow people to disable DEBUG mode, but set logging to
DEBUG. Now the UI will be nice and clean, users will have a small bit of
information which they can relay to an admin who can find the
corresponding information in the log.

### DIFF
--- a/crits/dashboards/views.py
+++ b/crits/dashboards/views.py
@@ -1,17 +1,33 @@
+import json
+
 from django.contrib.auth.decorators import user_passes_test
-from crits.core.crits_mongoengine import json_handler
-from crits.core.user_tools import user_can_view_data
+from django.core.urlresolvers import reverse
+from django.http import HttpResponse
 from django.shortcuts import render_to_response
 from django.template import RequestContext
-from crits.core.handlers import generate_global_search, generate_dashboard
-from crits.core.views import dashboard
-from django.http import HttpResponse
-from crits.dashboards.dashboard import SavedSearch, Dashboard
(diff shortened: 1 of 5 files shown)


mgoffin:
It doesn't include the username in the log for the person who generated the 500, but I wanted to get this PR out there so people can see what changes were made and also suggest things that might make it more useful without exposing too much information.

I also wound up cleaning up the dashboard views file because I used it to raise random Exceptions and there were tons of pyflakes errors :)

crits/dashboards/views.py
```diff
@@ -1,17 +1,33 @@
...
+from crits.dashboards.dashboard import Dashboard
+from crits.dashboards.handlers import (toggleTableVisibility,
+                                       get_saved_searches_list,get_dashboard,
+                                       clear_dashboard,
+                                       save_data,
+                                       get_table_data,
+                                       generate_search_for_saved_table,
+                                       delete_table,
+                                       getRecordsForDefaultDashboardTable)
+from crits.dashboards.handlers import (switch_existing_search_to_dashboard,
```
dsnellgrove: Ew giant whitespace blocks. Don't need to repeat the "from crits.dashboards.handlers import" which was just there again to break up the wide line of imported functions

dsnellgrove:
Yus, this is helpful. :+1:

crits/dashboards/views.py
mgoffin: Yep. I was just cleaning quick. This is much better than the single line that wrapped 4 times ;)

wxsBSD:
I'm testing this out and it works exactly as expected. I'm happy to merge this in whenever.

One more thought: we could use a similar thing for 404's, though a separate PR would be best.

mgoffin:
Sure!

PR #379, opened 2015-01-30
Fixed bug for missing password w/ldap

sgerlach:
When using LDAP auth with Active Directory, if a valid user is passed with no password, situations exist where AD will let a user bind with no password (anonymous bind) and success login is processed. Force a non-blank username and password.

### DIFF
--- a/crits/core/user.py
+++ b/crits/core/user.py
@@ -907,6 +907,10 @@ class CRITsAuthBackend(object):
         :returns: :class:`crits.core.user.CRITsUser`, None
         """
 
+        # Need username and password for logins, checkem both
+        if not all([username, password]):
+            return None
+
         e = EmbeddedLoginAttempt()
         e.user_agent = user_agent
         e.remote_addr = remote_addr


frbapolkosnik:
This might be going a bit deeper than that. There is a fallback feature that uses a locally stored password in case when the LDAP/AD is down. In some cases that local password could be blank (e.g. the user was created through create_unknown_user or the admin didn't specify the password for the user account). 

Can somebody using TOTP check if this patch is not causing any issues?

sgerlach:
For what it's worth, there are some other methods to use to check and make sure an non-anonymous connection was made, but it involves using ldap_whoami_s and comparing usernames or doing searches based on the DN. Both of which can get really messy depending on the implementation of ldap... (real v. AD)

wxsBSD:
I can confirm this doesn't break TOTP. I just tested it with a TOTP enabled account with no problems. I know nothing about LDAP, and especially how it is used in CRITs. If there is a rough consensus from the more LDAP-clueful people then I'll happily merge this in.

frbapolkosnik:
It works for me with my LDAP setup. But technically, in my opinion, the policy on the LDAP server should be in charge of enforcing password complexity requirements.

eltair:
I don't believe this is an issue of password complexity.  Many LDAP servers interpret a blank password as an anonymous bind even if the username is populated.  I have seen this issue in other applications and verified that it exists in crits. This needs to be fixed ASAP.  Relevant RFC snippet:

5.1.1.  Anonymous Authentication Mechanism of Simple Bind

   An LDAP client may use the anonymous authentication mechanism of the
   simple Bind method to explicitly establish an anonymous authorization
   state by sending a Bind request with a name value of zero length and
   specifying the simple authentication choice containing a password
   value of zero length.

eltair:
Even more relevant RFC section:

5.1.2.  Unauthenticated Authentication Mechanism of Simple Bind

   An LDAP client may use the unauthenticated authentication mechanism
   of the simple Bind method to establish an anonymous authorization
   state by sending a Bind request with a name value (a distinguished
   name in LDAP string form [RFC4514] of non-zero length) and specifying
   the simple authentication choice containing a password value of zero
   length.

   The distinguished name value provided by the client is intended to be
   used for trace (e.g., logging) purposes only.  The value is not to be
   authenticated or otherwise validated (including verification that the
   DN refers to an existing directory object).  The value is not to be
   used (directly or indirectly) for authorization purposes.

   Unauthenticated Bind operations can have significant security issues
   (see Section 6.3.1).  In particular, users intending to perform
   Name/Password Authentication may inadvertently provide an empty
   password and thus cause poorly implemented clients to request
   Unauthenticated access.  Clients SHOULD be implemented to require
   user selection of the Unauthenticated Authentication Mechanism by
   means other than user input of an empty password.  Clients SHOULD
   disallow an empty password input to a Name/Password Authentication
   user interface.  Additionally, Servers SHOULD by default fail
   Unauthenticated Bind requests with a resultCode of
   unwillingToPerform.

wxsBSD:
Am I correct in saying that there is still some disagreement from the various LDAP users and experts in this thread (of which I am neither)? I'm just trying to make those more clueful in this area can come to a rough consensus on if I should merge or not.

apolkosnik:
I've got no problem with the proposed change.

wxsBSD:
@eltair - As you've been the other recent voice in this and are clearly more clueful than I am with regards to LDAP, are you in agreement that this should be merged?

eltair:
I have applied this patch manually and verified that it does in fact prevent auth bypass when used with an LDAP server that supports anonymous binds.  I agree that it should be merged.

wxsBSD:
Thanks everyone!

PR #385, opened 2015-02-03
Stop duplicate MD5s created during bulk MD5 upload

brlogan:
If the same new MD5 is entered more than once in the Bulk Upload MD5 Samples table, and the table is uploaded, multiple Sample documents are created in the DB with the same MD5. These duplicate Samples cannot be accessed through the UI and must be manually merged/removed. This problem is caused by the caching optimization; the new Sample that was just added is not in the cache, so the code creates a duplicate Sample thinking that MD5 does not yet exist.

Fixed by updating the cache when a new Sample is added.


brlogan:
I'm not 100% certain this is the best solution, but it works.

mgoffin:
Seems decent enough for now. If there's better ways to do it we can update it again :)

PR #402, opened 2015-02-19
missing Ubuntu deps: libssl-dev swig + sudo

wzr:
The bootstrap scripts works mostly fine, it's just missing a couple sudos and 2 dependencies for Ubuntu 14.04.
I've added them in this PR.

Cheers.


mgoffin:
Thanks! We are looking to make more changes to this so the help is very much appreciated!

PR #407, opened 2015-02-23
Adding comment post api

jafilson:
Added enough features so that I may add comments via the API,  This is POST onlyl.


mgoffin:
The code looks good, haven't tested functionality. Looks like there's several imports that aren't used that can be cleaned up?

jafilson:
Possibly, I was unsure and just playing it safe when I left them there.

mgoffin:
It would be good to modify the help_main.txt file in the "documentation" directory to include the POST requirements for Comments. Make sure to add it alphabetically. Once that's added you can run `python rst2html.py help_main.txt help_main.html` from the "documentation" directory. That should update the HTML that you can view by going to "Help" in the Nav menu.

PR #408, opened 2015-02-23
Made changes to allow SHA1,SHA256,SIZE, and MIMETYPE

jafilson:
This allows you to add SHA1, SHA256, SIZE, and MIMETYPE when adding items with "METADATA" upload mode in the API.


mgoffin:
What are all of the changes to the relationship widget template for? They seem unrelated to the changes for this PR.

jafilson:
I forgot to split them out of this branch.   They create urls so that you can click on the items and go directly to them without the need to cut and paste

brlogan:
The relationship widget template already includes links to the related TLO. I'm not sure what you mean by the "need to cut and paste". Additionally, Indicator values are typically linked to a search for that value, not to the TLO itself.

mgoffin:
Can you do the documentation dance for this as well to include the new API POST parameters?

crits/samples/handlers.py
```diff
@@ -1016,6 +1026,18 @@ def handle_file(filename, data, source, method='Generic', reference=None, relate
     else:
         related_obj = None
 
+    if sha1_digest != None and sha1_digest != ""::
```
mgoffin: Double colon needs fixing.

crits/samples/handlers.py
```diff
@@ -1067,6 +1093,11 @@ def handle_file(filename, data, source, method='Generic', reference=None, relate
                                      "data itself, need to be supplied.")
                 retVal['success'] = False
 
+            if sha1_digest:
+                sample.sha1 = sha1_digest
+            if sha256_digest:
+                sample.sha256 = sha256_digest
```
mgoffin: Would you also want to add mimetype here?

jafilson: I didn't think it was needed here because its handled further up in the function: 

sample.mimetype = mimetype

mgoffin: Just checking since the hashes are also handled further up and wasn't sure why the mimetype and size weren't also added here.

crits/samples/handlers.py
```diff
@@ -1034,6 +1056,10 @@ def handle_file(filename, data, source, method='Generic', reference=None, relate
         sample = Sample()
         sample.filename = filename or md5_digest
         sample.md5 = md5_digest
+        sample.sha1 = sha1_digest
+        sample.sha256 = sha256_digest
+        sample.size = size
```
mgoffin: This should be cast to an int to make sure it doesn't go in as a string (https://github.com/crits/crits/blob/master/crits/samples/sample.py#L106)

mgoffin:
I tried uploading with a bogus sha1 and sha256 and a mimetype of "text/plain". It happily accepted the upload but the sha's and the mimetype are all missing.

I would expect an error on validation of the sha's or the acceptance of the mimetype if it's just going to drop the bad values on the floor.

brlogan:
Is this PR going to go anywhere? I'm interested in this functionality, specifically being able to add a SHA and SSDEEP hashes in the handle_file function. Do we just need to fix the casting to int, and error on validation issues?

mgoffin:
I made note of some issues in my previous comment and hadn't heard back from the developer. I guess if it's dead someone else can take the work as-is and finish it up and resubmit?

PR #409, opened 2015-02-27
Change sh to bash (bootstrap script and README)

ikoniaris:
Ubuntu's sh links to dash, and I was getting an error about the <<< operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Just run the following:
 
 ```bash
 
-    sh script/bootstrap
+    bash script/bootstrap
 ```
 
 ## Production CRITs install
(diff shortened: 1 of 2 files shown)


wxsBSD:
First, thanks for the suggestion! Rather than switch to assuming all the world has bash by default we should instead fix the code.

This seems to do the trick and is portable, though I'm far from a shell expert. Paging @sroberts!

```
wshields@crits-dev:~$ /bin/sh
$ . /etc/lsb-release
$ OS=$DISTRIB_ID
$ OS=$(echo $DISTRIB_ID | tr "[:upper:]" "[:lower:]")
$ echo $OS
ubuntu
$
```

mgoffin:
Keeping with sh and making ubuntu work properly would definitely be the preference.

ikoniaris:
Whatever works! It was mostly an FYI to be honest.

PR #417, opened 2015-03-12
Basic Validation and Domain Upload Fixes/Improvements

brlogan:
This started with a few Domain TLO upload issues and basic validation, and grew as I discovered a variety of additional issues. Here's a list of many of the areas this PR addresses:
- Basic validation of Domains, IPs, Email Addresses, and URLs
  - Goal: Limit user input errors, but still be fairly liberal on what is accepted
  - Covers single and bulk upload of Objects, Domains, IPs, and Indicators
  - Domains: Modified the get_domain function
    - Still validates TLD, but now also does not allow spaces or characters /:@\
  - IPs: Checks IPv4, IPv6, and CIDR IPs by themselves and in E-mail addresses and URLs
  - URLs: Applies Domain & IP checks to contents of URL, retains protocol prefix check
  - Email Addr: Checks for '@' and applies IP and Domain checks
  - Indicators: One function call to validate indicators
- IP Address Normalization: Strips leading zeros from IPv4 octets and uses the RFC 5952 canonical format for IPv6 addresses preventing duplicate entries when the same IP is entered in different ways
- Domain Upload Improvements:
  - IP field now accepts CIDR-notated IPs
  - Domain TLO is not added if included IP is invalid
  - Bulk Upload's "IPs Use Same Source As Domain" button now works
  - Bulk Upload table validation now helpfully displays errors for Object Data
  - Function "retrieve_domain" now queries the DB if item isn't found in the cache
  - Function "add_object" no longer returns "Object Added Successfully" upon successful validation
- Bulk IP Upload:
  - Source Method field is now set instead of always being "None"
  - Validation with an invalid IP now produces an error message and colors the bad row red
- Indicator upload:
  - Improved display of messages
- Function add_source now will not add a new source instance if an identical one (same method, reference, date, and analyst) already exists
-  Function add_ticket now supports EmbeddedTicket class objects as input
- Fixed a few domain upload issues with the new "create_indicator_from_obj" function
  - Renamed to "create_indicator_from_tlo"  to clairfy it is for TLOs and reduce confusion with the "create_indicator_from_object" function, which is for Objects
  - Multiple source instances were sometimes being created for the Domain and the Indicator
  - Only one campaign was inherited by the Indicator, now all are
  - Tickets were not inherited by the Indicator, now they are


mgoffin:
This is great! Haven't looked at anything yet but I'll hopefully get to it today. Have you looked at crits_services to make sure function name changes don't break things? I suspect add_indicator_from_obj was used in a couple services.

crits/domains/api.py
```diff
@@ -64,7 +64,7 @@ def obj_create(self, bundle, **kwargs):
         ip_method = bundle.data.get('ip_method', None)
         ip_reference = bundle.data.get('ip_reference', None)
         # Also add indicators
-        add_indicators = bundle.data.get('add_indicators', None)
+        add_indicator = bundle.data.get('add_indicator', None)
```
mgoffin: Hesitant to change the underlying API, but I can easily be knocked off the fence if other people want to see this change. I know the API is heavily used and this could cause problems for people who have automated systems up and running.

brlogan:
No, I had not looked at crits_services. In checking now, it looks like there are 3 html files that link to view "indicator_from_obj", and 1 instance of "get_domain" and "upsert_domain". I'll put in a PR to update those.

As for add_indicator(s), I initially removed the 's' because the New Domain form doesn't have the 's' and I wanted it to be more consistent.  But now, after taking a closer look, I've realized it being plural is probably more correct because an Indicator would be added for both the Domain and the associated IP. I think I'll go ahead and make add_indicator plural again and then we won't have the API change issue.

mgoffin:
Sounds good!

brlogan:
Okay! The services PR has been created and add_indicator changed!
I look forward to your other comments. :)

webstergd:
I think your additions are great and I wish you did it earlier. My only
question is in your modification you removed my VT check to see if a domain
is actually an IP. I have seen a few instances where VT and other services
will provide a domain in a form similar to http://102.32.234.3/stuff

So my question is, does your upsert modification check to see if the domain
is actually an IP?

brlogan:
Thanks, @webstergd!

Yes, upsert_domain now does the validation so you don't have to in advance.  upsert_domain will return `{'success': False, 'message': "<some message>"}` if the given domain:
- contains a space, slash, colon, or '@'
- Does not contain a valid TLD

Therefore, in your case, any URLs will be skipped, and IPs will be skipped (since they don't have a valid TLD). I believe it should function about the same as you originally had it, but please let me know if I broke something.

webstergd:
ohh that's beautiful.

wxsBSD:
I'm sure others know this but I'm putting it here just in case, and to make it a bit more formal. This is tied to https://github.com/crits/crits_services/pull/93. That one can not be merged until this is handled.

brlogan:
Thanks, Wes! I definitely should have linked those.

misterjulien:
Hopefully my issue is the same as one of the ones you're working on ...

I added an indicator for a Windows registry key and afterward none of my dashboards showed up.  After deleting the indicator, the dashboards showed up.  The registry key I added had "<" and ">" in it.  For example, "HKU\<GUID>\software\blah\".  I'm guessing that the GT and LT characters were fouling up the html.

It looks like you might be working on input validation already so I didn't create a new issue.

brlogan:
@misterjulien: This seems to me to be more of an issue with proper escaping of characters rather than input validation. The old style dashboard from 3.1 loads just fine with your given input, but I was able to reproduce your issue with the new custom dashboards.
Because angle brackets are valid input for some Indicators, I think this problem falls outside the scope of this particular PR. I recommend creating a new issue to track this.

mgoffin:
This will need a merge from master. I tried merging locally. There was one conflict in indicators/handlers with imports which was easy to clean but there are other places which call "get_domain" now which break since it was renamed.

brlogan:
Okay, it should now be ready to go!

PR #425, opened 2015-03-17
Add support for listing campaigns by name.

wxsBSD:
Add a view which calls into get_item_names() to fetch a list of
campaigns. It takes an optional flag to get active campaigns or inactive
campaigns.

While here, add a check in get_item_names() to make sure it can not be
used to enumerate sources.

### DIFF
--- a/crits/core/handlers.py
+++ b/crits/core/handlers.py
@@ -657,6 +657,10 @@ def get_item_names(obj, active=None):
     :returns: :class:`crits.core.crits_mongoengine.CritsQuerySet`
     """
 
+    # Don't use this to get sources.
+    if isinstance(obj, SourceAccess):
+        return []
+
     if active is None:
        c = obj.objects().order_by('+name')
     else:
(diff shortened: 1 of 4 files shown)


wxsBSD:
There is a corresponding pull request which uses this feature in https://github.com/crits/crits_services/pull/95. I would normally try to avoid adding a requirement on core for a service but this feature is useful beyond the service.

crits/core/handlers.py
```diff
@@ -657,6 +657,10 @@ def get_item_names(obj, active=None):
     :returns: :class:`crits.core.crits_mongoengine.CritsQuerySet`
     """
 
+    # Don't use this to get sources.
+    if isinstance(obj, SourceAccess):
+        return []
```
mgoffin: Didn't realize that this function is supposed to return a CritsQuerySet. By returning a list I don't know what that would do (I assume break stuff?).

wxsBSD: It doesn't break anything, but that is by virtue of the fact that nothing calls ```get_item_names(SourceAccess)```. Any place that wants to enumerate sources uses the corresponding handler from that. This was added more as an anti-foot-shooting measure in case someone screws up and calls ```get_item_names(SourceAccess)```.

PR #427, opened 2015-03-18
Standards Parser now uses Handlers

mgoffin:
For a long time we parsed STIX/CybOX and just created a new CRITs object or returned the current one in the database. The main issue with this is that it breaks the philosophy of running all of our data through handlers so no matter where you come from (web, API, command-line) you will be guaranteed that the same code is handling the same type of feature. This caused problems like creating Indicators but doing all of the automated fun stuff that would come along with creating an Indicator through the UI.

This change makes the Standards Parser parse out the necessary values and submit them as arguments into the appropriate handler for creating new content.

This might make the "from_stix()" or "from_cybox()" class methods deprecated, but I've left them in there for the time being until we are 100% sure we are good to remove them.

### DIFF
--- a/crits/objects/object_mapper.py
+++ b/crits/objects/object_mapper.py
@@ -66,6 +66,12 @@ class UnsupportedCRITsObjectTypeError(Exception):
     def __str__(self):
         return repr(self.message)
 
+def get_object_values(obj):
+    try:
+        return obj.values
+    except:
+        return [obj.value]
+
 def make_cybox_object(type_, name=None, value=None):
     """
     Converts type_, name, and value to a CybOX object instance.
@@ -322,20 +328,20 @@ def make_crits_object(cybox_obj):
     o.datatype = "string"
     if isinstance(cybox_obj, Account):
         o.object_type = "Account"
-        o.value = cybox_obj.description.values
+        o.value = get_object_values(cybox_obj.description)
         return o
     elif isinstance(cybox_obj, Address):
         o.object_type = "Address"
         o.name = str(cybox_obj.category)
-        o.value = cybox_obj.address_value.values
+        o.value = get_object_values(cybox_obj.address_value)
         return o
     elif isinstance(cybox_obj, API):
         o.object_type = "API"
-        o.value = cybox_obj.description.values
+        o.value = get_object_values(cybox_obj.description)
         return o
     elif isinstance(cybox_obj, Artifact):
         o.object_type = "Artifact"
-        o.value = cybox_obj.data.values
+        o.value = get_object_values(cybox_obj.data)
         if cybox_obj.type_ == Artifact.TYPE_GENERIC:
             o.name = "Data Region"
             return o
@@ -347,30 +353,30 @@ def make_crits_object(cybox_obj):
             return o
     elif isinstance(cybox_obj, Code):
         o.object_type = "Code"
-        o.value = cybox_obj.code_segment.values
         o.name = str(cybox_obj.type)
+        o.value = get_object_values(cybox_obj.code_segment)
         return o
     elif isinstance(cybox_obj, Disk):
         o.object_type = "Disk"
         o.name = str(cybox_obj.type)
-        o.value = cybox_obj.disk_name.values
+        o.value = get_object_values(cybox_obj.disk_name)
         return o
     elif isinstance(cybox_obj, DiskPartition):
         o.object_type = "Disk Partition"
         o.name = str(cybox_obj.type)
-        o.value = cybox_obj.device_name.values
+        o.value = get_object_values(cybox_obj.device_name)
         return o
     elif isinstance(cybox_obj, DNSQuery):
         o.object_type = "DNS Query"
-        o.value = cybox_obj.question.qname.value.values
+        o.value = get_object_values(cybox_obj.question.qname)
         return o
     elif isinstance(cybox_obj, DNSRecord):
         o.object_type = "DNS Record"
-        o.value = cybox_obj.description.values
+        o.value = get_object_values(cybox_obj.description)
         return o
     elif isinstance(cybox_obj, DomainName):
         o.object_type = "URI - Domain Name"
-        o.value = cybox_obj.value.values
+        o.value = get_object_values(cybox_obj.value)
         return o
     elif isinstance(cybox_obj, EmailMessage):
         o.object_type = "Email Message"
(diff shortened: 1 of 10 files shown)


wxsBSD:
I took a quick glance at this and it appears right, and also moves things in the correct direction of using handlers for this. I'm not in a position to test it but it is my understanding that it is being actively tested by some people. My hope is that they will be able to chime in here and provide feedback.

crits/objects/object_mapper.py
```diff
@@ -322,20 +322,20 @@ def make_crits_object(cybox_obj):
...
     elif isinstance(cybox_obj, Address):
         o.object_type = "Address"
         o.name = str(cybox_obj.category)
-        o.value = str(cybox_obj.address_value)
+        o.value = cybox_obj.address_value.values
         return o
     elif isinstance(cybox_obj, API):
         o.object_type = "API"
-        o.value = str(cybox_obj.description)
+        o.value = cybox_obj.description.values
```
brlogan: I was doing some testing and encountered an error:
`API: 'StructuredText' object has no attribute 'values'`
It turns out that the description field of CybOX objects like "API", "Code", and "DNS_Record" (I'm sure there are more) are based on class StructuredTest which doesn't have a "values" property. Changing "values" to "value" seems to help, but I'm not certain that is the correct answer.

crits/standards/parsers.py
```diff
@@ -160,45 +253,179 @@ def parse_observables(self, observables):
...
-                obj = cls.from_cybox(obs)
-                obj.add_source(self.source)
-                obj.save(username=self.source_instance.analyst)
-                self.imported.append((cls._meta['crits_type'], obj)) # use class to parse object
+                if isinstance(item, Address):
+                    if item.category in ('cidr', 'ipv4-addr', 'ipv4-net',
+                                         'ipv4-netmask', 'ipv6-addr',
+                                         'ipv6-net', 'ipv6-netmask'):
+                        imp_type = "IP"
+                        for value in item.address_value.values():
```
brlogan: This is causing error `Address: 'list' object is not callable`.
It looks like the parentheses need to be removed from the end:
`for value in item.address_value.values:`
Same for the domains below.

crits/standards/parsers.py
```diff
@@ -160,45 +253,179 @@ def parse_observables(self, observables):
...
+                    if item.header:
+                        data['message_id'] = str(item.header.message_id)
+                        data['subject'] = str(item.header.subject)
+                        data['sender'] = str(item.header.sender)
+                        data['reply_to'] = str(item.header.reply_to)
+                        data['x_originating_ip'] = str(item.header.x_originating_ip)
+                        data['x_mailer'] = str(item.header.x_mailer)
+                        data['boundary'] = str(item.header.boundary)
+                        data['from_address'] = str(item.header.from_)
+                        data['date'] = item.header.date
```
brlogan: This line is leading to error: 
`Observable: Failed to save object. <br /><pre>'DateTime' object has no attribute 'read'</pre>`

`item.header.date` is a `cybox.common.properties.DateTime object`. Because it isn't a standard datetime.datetime, the CRITs Email class "_custom_save" method tries to use dateutil.parser, but it chokes.
Changing `data['date'] = item.header.date` to `data['date'] = item.header.date.value` seems to help, but again, I'm not certain that is the best solution.

mgoffin:
Thanks @brlogan! I've made some adjustments. I'm not a huge fan of the way I did it, but it might allow for us to be "lazy" about the attribute types being returned and "flexible" enough to accommodate them being altered in the future. Do these changes clean up any of your problems?

brlogan:
Thanks for fixing those up so fast, @mgoffin! I think your solution should work well and it appears to have solved the bugs I pointed out. Between other work over the last couple days, I managed to figure out another bug I was seeing. I'll comment on the line to describe it and you can let me know what you think.

crits/standards/parsers.py
```diff
@@ -160,45 +253,179 @@ def parse_observables(self, observables):
...
+                    self.parse_res(imp_type, obs, res)
+                elif isinstance(item, File):
+                    imp_type = "Sample"
+                    filename = str(item.file_name)
+                    md5 = item.md5
+                    data = None
+                    for obj in item.parent.related_objects:
+                        if (isinstance(obj.properties, Artifact) and
+                            obj.properties.type_ == Artifact.TYPE_FILE):
+                            data = base64.b64decode(obj.properties.data)
```
brlogan: I was seeing a `File: Incorrect padding` failure message, but only when receiving certain Samples via TAXII Agent. I eventually realized that even the Samples that were making it through without an error had bad data. After some work, I managed to trace this problem to the above line.  I didn't dig very deep into the python-stix code, but it appears that `STIXPackage.from_xml()` does the work of base64 decoding the Sample data. So, when the above line executes, it is trying to decode already decoded data, which causes the error.
Changing `data = base64.b64decode(obj.properties.data)` to `data = obj.properties.data` seems to solve the problem.

mgoffin:
I can fix that up too! One thing that isn't working because I left it as a TODO is email attachments. If a STIX doc comes in for an email and it has an attachment, just the attachment gets dropped on the floor. Should probably figure out how to get the attachment out and DTRT. I wonder if it will have the same b64decode situation.

crits/standards/parsers.py
mgoffin: I wonder if it does the same things for PCAP, RawData, and Certificate binary data...

packet-rat: @mgoffin:  I wonder if it does the same things for PCAP: Yes

voila_capture 2015-03-26_09-31-54_am

mgoffin: @packet-rat That's a completely unrelated error, but I'll fix that too :)

mgoffin:
Ok so I think I got that fixed. I was able to test a Sample, PCAP, and RawData and they all got added correctly. I nixed the description stuff since these objects don't have description attributes, and fixed a bug with RawData where we were trying to encode a string causing decode problems.

brlogan:
@mgoffin: Have you yet done any work towards including email attachments? I've looked into it a bit and come across a potential issue with python-cybox. I'll CC you on my email to Greg. If you've already made progress on this, please tell me to poke elsewhere.

mgoffin:
I haven't made any progress, but I'm going to keep an eye on where that email thread goes! Granted it's in the other direction, but it might clue me in on how to get the attachment back out :)

mgoffin:
If we are happy enough where this has gotten to and are good with fixing any more bugs as we find them, I'm ok with merging this into master :)

The email attachment thing is going to take some time and the bug that @brlogan found isn't in CRITs.

brlogan:
I think we're pretty happy with the things this PR was intended to address. I'm fine with merging this to master.

If it's fine with you, @mgoffin, I'll open up an Issue to track the STIX email attachment situation.

mgoffin:
@brlogan sounds good to me!

Issue #383, opened 2015-02-03
Updating TLOs via the API


mgoffin:
Design discussion so far:

* Manage a list of "actions" that can be taken somewhere in the core API:
  * Add/edit/delete a Ticket, Campaign, Comment, etc.
* Map "actions" to appropriate handlers.
  * If an action is provided we don't support, respond with appropriate error.
  * If we find the appropriate action, pass along POST data to the handler (see below).
* Manage necessary POST data for each "action".
  * Includes a section in the documentation.
  * POST data should match directly to the expected arguments for the associated handler so we can do something like `handler(**post_data)`.
  * Assign default values for all arguments and add "*kwargs" to each handler so any extra POST data the user tries to pass in can be safely ignored and not throw errors.
    * This makes the API a bit more resilient to handler argument changes.

The goal here is to prevent writing any extra code in the API to interface with the handlers. That way any changes we make to the handlers will work appropriately with the API.

We should probably use PATCH for this (http://django-tastypie.readthedocs.org/en/latest/interacting.html#partially-updating-an-existing-resource-patch).

PR #430, opened 2015-03-23
Initial commit for new delete approach

puhley:
I am creating a new branch and pull request focused solely on deleting top level objects. Based on the discussion around #333 and #383, I think it will be easier if we separate the discussion and code for the DELETE calls from the discussion and code for the POST/PATCH calls.

This code is not yet complete. I would like the design to be reviewed before finalizing it. In this design, deletion for any TLO can be handled by the core api.py. The core api.py file uses the core handler.py to delete the object. The handler.py has been modified so that the same deletion code can be used for web requests and API requests. Even though the core api.py has code to delete any TLO, you can't delete a specific TLO type until 'delete' is added to the list of allowed methods within that TLO's api.py file. If a particular TLO needs special handling, as is the case with campaigns, then that TLO's api.py file can override the core code with it's own implementation. This should allow for the code to be centralized, allow for individual delete permissions for each TLO, and allow a TLO to override the default deletion code when customized actions are needed.

Let me know if this is closer to what you had in mind and if you have any feedback for improvements.

### DIFF
--- a/crits/campaigns/api.py
+++ b/crits/campaigns/api.py
@@ -2,11 +2,25 @@ from django.core.urlresolvers import reverse
 from tastypie import authorization
 from tastypie.authentication import MultiAuthentication
 from tastypie.exceptions import BadRequest
+from bson.objectid import ObjectId
 
 from crits.campaigns.campaign import Campaign
-from crits.campaigns.handlers import add_campaign
+from crits.domains.domain import Domain
+from crits.emails.email import Email
+from crits.events.event import Event
+from crits.indicators.indicator import Indicator
+from crits.ips.ip import IP
+from crits.pcaps.pcap import PCAP
+from crits.samples.sample import Sample
+from crits.raw_data.raw_data import RawData
+from crits.actors.actor import Actor
+
+from crits.campaigns.handlers import add_campaign, campaign_remove, remove_campaign
 from crits.core.api import CRITsApiKeyAuthentication, CRITsSessionAuthentication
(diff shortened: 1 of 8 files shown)


mgoffin:
I think we can go in a bit of a different direction here. The primary thing of concern is that there is no proper leveraging of existing handlers for deleting TLOs. Code is duplicated here meaning that, for example, if we change the way we delete Indicators in `indicators/handlers.py` then the way an Indicator is deleted through the API is not the same.

All the API should do is be a light wrapper around all of the handler files and do exactly what all of the views do: take in data, massage it into the appropriate formats, and pass it to the handlers.

puhley:
OK. I misunderstood your original comments around code duplication which is why I made everything use the core API and the core handler. So each TLO api.py would have a lightweight delete_detail method which would call the corresponding deletion method in the TLO's handler.py?

mgoffin:
Basically we would do this:

* Modify the core `api.py` file and build out `obj_delete` `obj_delete` will do a couple simple things:

1) determine the TLO type being deleted
2) determine the ObjectId of the TLO being deleted
3) determine the user requesting the TLO be deleted
4) call the appropriate handler and pass in the appropriate information

That's basically it. Then we go through and enable the "delete" option for each TLO's API. If a TLO doesn't have a handler for deleting then we shouldn't enable its "delete" API option until one is created. If a TLO's delete handler isn't sufficient and thorough enough we shouldn't enable its "delete" API option until it is thorough enough.

PR #432, opened 2015-03-24
Updated handling of embedded sources to resemble what's happening in sam...

frbapolkosnik:
Updated handling of embedded sources in Raw Data to resemble what's happening in the handler for samples.

 handle_raw_data_file() in raw_data/handlers.py is not handling embedded_source objects.
I found that case when I was trying to add a text file produced by a tool that processed some sample data.

Calling hanfle_raw_data_file() with source derived from the sample (obj.source) would not work, and I'd get an object inthe database that nobody can ever see...

 res = handle_raw_data_file(data, self.obj.source, self.current_task.username,
                        title="Some Title", data_type='text',
                        tool_name='Tool name', tool_version='0.6', tool_details='http://www.nowrap.de/flare.html',
                        method=self.name,
                        copy_rels=True)

I was able to get a single source by either using a hardcoded string or by using self.obj.source[0].name

 res = handle_raw_data_file(data, self.obj.source[0].name, self.current_task.username,
                        title="Some Title", data_type='text',
                        tool_name='Tool name', tool_version='0.6', tool_details='http://www.nowrap.de/flare.html',
                        method=self.name,
                        copy_rels=True)

Then I looked into the handler for sample files, and I tried to replicate the same behavior, and now it all works with my service.

PR #438, opened 2015-03-31
RawData embedded sources handling fix

frbapolkosnik:
This updates the RawData's handle_raw_data_file() to accept EmbeddedSources. Before it would only accept a strings, and the embeddedSources would get written into the database and resulting in a RawData objects that are not visible to anyone. I've essentially reused the code from Samples' handle_file() to fix this.


frbapolkosnik:
This is what was done in https://github.com/crits/crits/pull/432

wxsBSD:
Much thanks, merging in!

PR #439, opened 2015-03-31
Relationship list UI change

frbapolkosnik:
This is just a visual change to make browsing of the related documents more intuitive, also a tool tip is added.

### DIFF
--- a/crits/emails/templates/email_listing_mini_widget.html
+++ b/crits/emails/templates/email_listing_mini_widget.html
@@ -20,6 +20,7 @@
     {% endif %}
         <thead>
             <tr style="border-top: 0px;">
+                <th>Details</th>
                 <th width="120px">Relationship</th>
                 <th>From</th>
                 <th>Sender</th>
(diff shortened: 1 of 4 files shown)


wxsBSD:
I'm curious what @mgoffin thinks of this. It does make it more in-line with how other "click here to go to the document" links work, which I think I'm in favor of. I'd be happy to merge this but I try to shy away from UI changes without further input from the community.

mgoffin:
I'll be honest, I don't like the implementation of this change. It's almost tripled the height of some of the rows that contain more columns (Events for me are the issue but I'm sure Samples will have the same problem).

wxsBSD:
But you are generally in favor of the details icon being used instead of the arrow? If Adam can find a way to keep things sane from an implementation perspective?

mgoffin:
I'd be fine if the change was just "make the arrow the Details icon from the jtable layout".

wxsBSD:
That is also fine by me. The arrow always struck me as "everything else uses the details icon, why this arrow?" @frbapolkosnik any issues with making these suggested changes?

frbapolkosnik:
One of the reasons for moving the icon to the left is that it requires much less travel when browsing through and that way it is more in line what's showing in the object listings (the first thing is the details icon). The other reason was that the relationship detail, relationship delete, and that was all bunched up at the end on the right side; on a large screen I had issues trying not to hit the trashcan icon instead.

 I bet that it all can be made nicer and tighter,

mgoffin:
@wxsBSD The arrow was put in place *long* before we ever had jtables and this "Details" icon :)

@frbapolkosnik If you want to just move the icon to the left and be the first column, remove the "Details" text so the column isn't unnecessarily huge (just put the icon in the header and have an alt text of "Details" for people who hover over it), and shrink the width of the last column so it matches the two remaining icons, it will probably do the trick.

apolkosnik:
Sounds good

brlogan:
I also support making this more consistent, but agree that we shouldn't sacrifice vertical space by making the relationships taller.

apolkosnik:
I got it, even managed to make it all much tighter than the original, but
it will have to wait until Monday.

@wxsBSD <https://github.com/wxsBSD> The arrow was put in place *long*
before we ever had jtables and this "Details" icon :)

@frbapolkosnik <https://github.com/frbapolkosnik> If you want to just move
the icon to the left and be the first column, remove the "Details" text so
the column isn't unnecessarily huge (just put the icon in the header and
have an alt text of "Details" for people who hover over it), and shrink the
width of the last column so it matches the two remaining icons, it will
probably do the trick.

—
Reply to this email directly or view it on GitHub
<https://github.com/crits/crits/pull/439#issuecomment-89338552>.

apolkosnik:
#447 Should address the comments. @mgoffin I hope this is what you had in mind. I'll address any pixel level issues on Monday. Thanks for the pointers.

PR #447, opened 2015-04-04
Relationship list UI change v2

apolkosnik:
This is just a visual change to make browsing of the related documents more intuitive, also a tool tip is added.
I've also tweaked Relationship_type_edit column as it was forcing everything around it to take more space vertically.

Since I don't have a CRITs instance handy at the moment, and I had a moment, this should be probably pretty close to what was requested in https://github.com/crits/crits/pull/439

### DIFF
--- a/crits/emails/templates/email_listing_mini_widget.html
+++ b/crits/emails/templates/email_listing_mini_widget.html
@@ -20,6 +20,7 @@
     {% endif %}
         <thead>
             <tr style="border-top: 0px;">
+                <th style="width: 16px; padding: 0px" class="ui-icon ui-icon-document" title="Details"></th>
                 <th width="120px">Relationship</th>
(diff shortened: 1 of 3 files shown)


mgoffin:
There were a couple issues with this as well:

- The icon in the header was doing weird border things because it wasn't encased in a span.
- The space for the Notes and Delete icon was shrunk too much so it caused them to wrap and take up more vertical space than necessary.
- The entire block that was moved for determining the URL was originally poorly written and could be collapsed down to a single line.

My OCD kicked in and I had to fix all of those and then I found a few annoying things in the templates while poking around. I wound up doing this real quick as it would be faster for me to just fix this than to try and go back and forth with the PRs to get things where they need to be.

Check it out in master and see what you think!

PR #448, opened 2015-04-04
Make descriptions part of CritsBaseAttributes.

wxsBSD:
We've heard multiple requests to make description a field on all
objects, and this will address that.

Put description in CritsBaseAttributes, which most objects inherit from,
and remove it from those objects which no longer need it.

Update all the various templates to use a description_widget template,
which handles everything necessary for descriptions, including the AJAX
updating.

There is now a generic view and handler for updating description across
all objects. The view is update_object_description() and the handler is
description_update().

There are only two minor special cases here:

The first special case is screenshots, which are handled in their own
javascript because descriptions are editable only in the listing page.
They also do not inherit from CritsBaseAttributes so they define their
own description attribute.

The second one is that Event objects have a required description, so
specify that eventhough it is inheriting from CritsBaseAttributes too.

While here, fix some minor display problems in templates that were using
extra "tr" tags.


wxsBSD:
Um, I think I need to make this honor source restrictions. Fix coming later tonight.

crits/core/templates/description_widget.html
```diff
@@ -0,0 +1,47 @@
+<td class="key">Description</td>
+<td class="longtext">
+    <span class="edit_underline" id="object_description">{{description}}</span><span id="object_description_error"></span>
+</td>
+<script type="text/javascript">
```
mgoffin: This should all go into a .js file, not sit in the template.

wxsBSD:
Addressed your concerns @mgoffin.

PR #451, opened 2015-04-08
updated bootstrap and requirements to work on freebsd 10+

kai5263499:


mgoffin:
Excellent! I know someone who would be interested in this :) I'll try to do some testing on non-FreeBSD hosts to make sure this doesn't alter them as well. Might be a day or two.

frbapolkosnik:
Wes,

Can you please test change I made in  #479, if it still works for you.
It seems that the docs show an example where the one-liner that you added should be split in two lines or the pip ignores the version requirement for mongoengine causing havoc and confusion ;-).
Thanks,
Adam

PR #452, opened 2015-04-09
Enable Cross Site Request Forgery (CSRF) protection.

gbartz:
This change enables Cross Site Request Forgery (CSRF) protection.
https://docs.djangoproject.com/en/1.6/ref/contrib/csrf/

Enable django middleware in settings.py:
'django.middleware.csrf.CsrfViewMiddleware'

Template forms get {% csrf token %} where required.

AJAX forms get csrf token added to the header in $.ajaxSetup()
                Read the cookie
                Add token to header
                
Non-AJAX forms are handled individually as I tested.
                Read the cookie
                Add token as hidden input in the form
                
Dynamic javascript forms
                Read the cookie
                Add hidden input as string before form is appended to the body

### DIFF
--- a/crits/core/static/js/core.js
+++ b/crits/core/static/js/core.js
@@ -646,6 +646,19 @@ function initTabNav() {
     });
 }
 
+var csrftoken = readCookie('csrftoken');
+function csrfSafeMethod(method) {
+    // these HTTP methods do not require CSRF protection
+    return (/^(GET|HEAD|OPTIONS|TRACE)$/.test(method));
+}
+$.ajaxSetup({
+    // Set request header for ajax POST
+    beforeSend: function(xhr, settings) {
+        if (!csrfSafeMethod(settings.type) && !this.crossDomain) {
+            xhr.setRequestHeader("X-CSRFToken", csrftoken);
+        }
+    }
+});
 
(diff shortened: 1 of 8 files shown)


dsnellgrove:
If you find yourself getting a 403 from something after accepting this it's probably the CSRF fault. CRITs is kind of all over the map with how it implements sending requests to the web server so we patched all the ways and items we could find (html forms, js-created forms, ajax requests) but may have missed some.

mgoffin:
I got a 403 after my session timed out and i attempted to log in. But after I refreshed that issue went away and I can't reproduce with the Logout feature so I think it was just a weird fluke.

Going to accept this and we can handle 403's that pop up as needed (I suspect some services might have a problem).

frbapolkosnik:
My instance is totally useless right now, I'm getting 403 for everything, no request can get through,

brlogan:
I am also seeing 403's whenever I do a POST, but I haven't had time to do any investigation.

mgoffin:
What web server are you running? I tested with runserver and posted some new indicators and such after logging in and it was fine.

brlogan:
This is with runserver. If I log out, I can't do a POST to log back in.

Issue #446, opened 2015-04-03
Remove "Campaigns" section in favor of Relationships


mgoffin:
Now that Relationships have a "confidence" and "reason" there is basically no reason to keep the "Campaigns" section on the Details page around as a feature. We should use Relationships to relate a TLO to a Campaign.

One issue with this is that Relationships are a bit cumbersome to add. One would have to go to a TLO, copy the ID to the clipboard, then go to the Campaign and add the relationship.

What I  propose is that we do the following:

- Remove the "campaigns" TLO attribute.
  - Which means we can remove the Embedded Campaign document stuff as well as one of the classes that gets inherited for adding campaign attribution to a TLO.
- Migrate existing Attribution to Relationships using migrate.
  - Also add this to the "upgrade" management command to make this easy for people to do.
- Remove the "Campaigns" section of the Details page.
- Remove all of the tabs on the Campaign Details page for the different TLOs.
- Add a button to the top-right (near Favorites) which is "Attribute Campaign".
  - Or make it another button next to the "Add Relationship" button?
  - The button would bring up a form similar to the current "Add Campaign" button.
    - We would ask for a:
      - Campaign (populated dropdown list) - Required
      - Relationship Type (populated dropdown list, with "Related To" as the default) - Required
      - Confidence (populated dropdown based on the relationship confidence values) - Required
      - Reason (open text field) - Optional
  - The form would submit but would now add a Relationship and updated the "Relationships" section of the Details page automatically.
- Remove the Campaign mapreduce since it will no longer be necessary.
- Update the Campaign listing page so that it shows counts of each TLO that it is related to.
- Update the listing pages of other TLOs which currently show the list of Campaigns so that it gets this information from Relationships.
- Update the "Relationships" service to remove the special need to check the "campaigns" attribute, and let it run completely off of Relationships.

This would also wind up fixing #443.

Anything else I might have missed trying to make this work?

thelok:
So... my investigation into relationships back a while ago.

The code on the back end for loading relationships is not efficient. Let's say you have 1000 domains attributed to a campaign. Fine, each domain will have a single relationship to the campaign. But open up the details page for the campaign and it will have 1000 relationships -- it should load take a long time to process. It does some sorting and I forget where that code was located...

As long as testing is done with scale in mind and the UI is still easy to use then I don't have complaints.

mgoffin:
If anyone has any ideas for how to make it more efficient i'm all about it :)

brlogan:
[...]

PR #456, opened 2015-04-10
Improve Target Visibility and Handling

### LINKED ISSUE #443: Problem adding targets to campaign
There's a problem with the expected functionality of the "Targets" object and their association with campaigns.  I would expect the ability to add a "target" to a campaign regardless of whether or not I have imported any email into CRITs to this campaign.  As it currently exists, "Targets" only show up in a campaign if I have added a target to the campaign and imported an email targeting the user that is also associated with the campaign.

brlogan:
I was seeing a variety of problems when the same email address, but with different case, exists in multiple Targets.  Ideally we could consider email addresses case-insensitive and an address would only appear once in the Targets collection. While the Email TLO would retain record of how the address was actually cased, the Targets collection would consist of only lower-cased email addresses. However, this is not necessarily the case.

I made a few changes to **Targets** to address that and a couple other issues:
* Adding/Updating:
  * Problem: Using “iexact” in upsert_target was sometimes causing the wrong Target to be updated. If a user updates a Target, but the same email address with different case also exists as a Target, the other Target may be updated instead.
  * Fix: Now the code will first look for an exact match, if none is found it will look for a case-insensitive match (to avoid creating a case-insensitive duplicate). If still none is found, a Target will be created with a lower-case version of the email address.

* Deleting
  * Problem: Due to the use of “iexact”, sometimes the wrong Target would be deleted.
  * Fix: Only an exact match is deleted.

* Details
  * Problem: Using “iexact” here makes it so the user may not be able to access the Target they intend to view.
  * Fix: When requesting a Target details page, because the email address is used as the unique identifier, do the following:
    * First look for an exact email address match
    * If an exact match isn’t found, find a case-insensitive match.
    * If a match still is not found, use the lower-case version of the email address.
    * Emails tab and count uses a case insensitive query

* Adding to Campaign
  * Change: If the user is viewing Campaign Details and clicks the “Add Target” button at the top of the Targets jtable, the form will automatically select the current Campaign.
  * **Note**: I’m pretty sure there is a better way to do the javascript for this, so I’m open to suggestions. :)

* Listing in Campaign
  * The jtable in the Campaign's "Targets" tab will include:
	* Every campaign-attributed Target (Fixes #443)
	* Every Target with an email address matching an exact, or lower-cased, address within the “To” line of an attributed Email.
		* The “correct” Target will contain the lower-cased version of the email address.
		* The exact email address is included so that the user is made aware of old or duplicate Targets that weren’t lower-cased.
		* I would have liked to do a case-insensitve query here so that the user would be made aware of any duplicate Target, but I didn't for performance reasons.
  * Email Count uses a case insensitive query. Differently cased version of the same email address will have the same count.

**Note**: I realize some of this will be affected by the potential upcoming Campaign changes (#446), but since that might not happen for awhile, I thought it would be good to address these things now.

### DIFF
--- a/crits/campaigns/handlers.py
+++ b/crits/campaigns/handlers.py
@@ -26,6 +26,7 @@ from crits.indicators.indicator import Indicator
 from crits.ips.ip import IP
 from crits.pcaps.pcap import PCAP
 from crits.samples.sample import Sample
+from crits.targets.handlers import get_campaign_targets
 from crits.targets.target import Target
 
 
@@ -92,19 +93,7 @@ def get_campaign_details(campaign_name, analyst):
                                                               __raw__=formatted_query).count()
 
     # Item counts for targets
-    emails = Email.objects(source__name__in=sources, __raw__=formatted_query)
-    addresses = {}
-    for email in emails:
-        for to in email['to']:
-            # This might be a slow operation since we're looking up all "to"
-            # targets, could possibly bulk search this.
-            target = Target.objects(email_address__iexact=to).first()
-
-            if target is not None:
-                addresses[target.email_address] = 1
-            else:
-                addresses[to] = 1
-    uniq_addrs = addresses.keys()
+    uniq_addrs = get_campaign_targets(campaign_name, analyst)
     counts['Target'] = Target.objects(email_address__in=uniq_addrs).count()
 
     # favorites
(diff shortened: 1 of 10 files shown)


brlogan:
Added commit to enable global search of Target email address and name.

PR #459, opened 2015-04-14
Locations

mgoffin:
Adding locations feature to CRITs. This is a TLO feature (available for all TLOs) that tracks where that TLO has been observed originating from or destined to. This can mean many things depending on the TLO. For example:

* Geo-location for src and dest IP in a PCAP flow
* Email originating server and destination server
* Location of where a sample was downloaded from, or where the machine was that it was downloaded to.

Locations are unique based on type, location, and date of entry. Each location comes with an optional description, latitude, and longitude that can be modified afterwards if necessary. Location is based on a list of countries/territories.

Using this requires to run the following:

```
python manage.py create_locations
```


mgoffin:
Any C&C on this before we move forward?

PR #470, opened 2015-04-28
Fix the CSRF for toast notifications

frbapolkosnik:


thelok:
+1, Thanks

mgoffin:
Why do we need this here when we have this?

https://github.com/crits/crits/blob/master/crits/core/static/js/core.js#L649

mgoffin:
The other function also exists in core right under the last function.

frbapolkosnik:
I have to apologize, it looks like the whole issue was due to some stale cached javascript.

PR #474, opened 2015-04-29
Indicators Jtable Listing Deletes Columns on Refresh

thelok:
Need to shallow copy the jtopts_fields field otherwise the actual list will be deleted and persisted. If you keep refreshing the jtable listing for indicators then a new column will be deleted each time.

HO HO HA HA


mgoffin:
lol, i suck :) thanks!

PR #480, opened 2015-05-01
Couple of small changes to make Screenshots more usable.

apolkosnik-old:
At first, I've added the same kinds of changes that I made for raw_data in order to propagate source from the parent sample. 

Then I added timestamp for the embedded source creation. 

I also changed s.save() to s.save(username=analyst)
and added two calls to .reload() after data is saved.

Please take a look and comment if you'd feel that I'm doing something the wrong way.

Thanks!

### DIFF
--- a/crits/screenshots/forms.py
+++ b/crits/screenshots/forms.py
@@ -28,7 +28,7 @@ class AddScreenshotForm(forms.Form):
                            help_text='Comma-separated list of tags')
     source = forms.ChoiceField(required=True,
                                label=form_consts.Object.SOURCE,
-                               widget=forms.Select(attrs={'class': 'no_clear bulknoinitial'}))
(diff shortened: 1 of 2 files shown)


mgoffin:
When adding a screenshot to a TLO it does not inherit the sources which I would expect to be the behavior since you're noting that it will propagate sources from the parent.

frbapolkosnik:
@mgoffin do you need that for the AddScreenshot part in the UI when adding the screenshots by hand or just take the embeddedSource from the TLO and shove it into the Screenshot?

mgoffin:
Given that a screenshot can be associated with multiple TLOs, you'd have to merge any duplicate sources so all of the instances stack up properly as well as add any new sources?

frbapolkosnik:
I'll have to come up with a few test cases for this.

apolkosnik-old:
This is no longer needed after #896, since it was almost the same thing.

PR #485, opened 2015-05-05
New TLO from selected text

mgoffin:
When you select text a small floating menu will pop up allowing you to create a new Indicator, Domain, or IP from the value you selected. All it does is open up the form and insert the selected value into it.

Could be expanded in the future to do more stuff I guess, but at least the framework for it is there. I can think of some pretty neat stuff to do with selected text but this seemed like the most obvious.

### DIFF
--- a/crits/core/static/js/core.js
+++ b/crits/core/static/js/core.js
@@ -660,7 +660,50 @@ $.ajaxSetup({
     }
 });
 
+var selected_text = null;
+
+function getSelected() {
+    if (window.getSelection) {
+        return window.getSelection();
+    } else if (document.getSelection) {
+        return document.getSelection();
+    } else {
+        var selection = document.selection && document.selection.createRange();
+        if (selection.text) {
+            return selection.text;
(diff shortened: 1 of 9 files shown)


landw1re:
Mike does this functionality work with results produced by services? I've been looking for something to quickly add indicators based off of results that are produced from services like Cuckoo, etc. This is an excellent addition and allows for quick and fluid workflow.

I'm currently tracking stable_4. If I were to cherry pick into stable_4 would you foresee any significant issues? Also, when this code is fairly stable can I make the request to push it to stable_4 :-) ?

Again, great addition.

mgoffin:
The functionality works (or should work!) on any page in CRITs no matter where it is or what renders it.

I would recommend testing it in master and not cherry-picking it to stable_4. It will land there eventually once it's accepted and until then there could be changes if people find issues or have recommendations.

landw1re:
Thanks Mike. I'll definitely check it out using master then.

PR #492, opened 2015-05-06
Backdoor and exploit

wxsBSD:
Implement backdoor and exploit TLO.

These are intended to replace the backdoor and exploit attributes of Sample objects. By breaking them out into their own TLO we get more flexibility in relating, grouping and tracking these.

I have provided all the necessary bits to migrate existing backdoor and exploit attributes into their own objects.


wxsBSD:
I'll add an UPDATING entry and also API docs tonight.

wxsBSD:
One thing to keep in mind for anyone looking at this. When you add a backdoor via the handler it can create up to two different objects. The first object is what I call the "family" object, which is a backdoor name with no version. The second object is what I call the "specific backdoor" object, which is a backdoor name with a version.

This is useful because it allows you to relate things to the family if you don't know (or don't care) about the version, but also if you care to get into specific version tracking you can do that too. The only time we create a specific backdoor object is if you provide a version when creating the backdoor.

jezkerwin:
So does this mean that Backdoor and Exploit will appear as an option in the main menu, like for Domains or IPs?
Am I understanding a TLO correctly?
:-)
Cheers.

wxsBSD:
Yes. TLO stands for "top level object" in CRITs speak.

crits/backdoors/handlers.py
```diff
@@ -0,0 +1,423 @@
...
+    :returns: dict with keys:
+              "success" (boolean),
+              "message" (str),
+              "id" (str),
+              "object" (if successful) :class:`crits.backdoors.backdoor.Backdoor`
+    """
+
+    retVal = {'success': False, 'message': ''}
+
+    if isinstance(source, basestring):
```
mgoffin: Might add an elif here in the event source is an EmbeddedSource object in which case you need to make a list as well. Currently only handles string and list for source.

crits/backdoors/handlers.py
```diff
@@ -0,0 +1,423 @@
...
+        family = Backdoor()
+        family.name = name
+        family.version = ''
+        # Family does not exist, new object. Details are handled later.
+        objs.append(family)
+    else:
+        # Family exists, old object. Add a new source only.
+        for s in source:
+            family.add_source(s)
+        family.save()
```
mgoffin: Not passing in the user to .save()

crits/backdoors/handlers.py
```diff
@@ -0,0 +1,423 @@
...
+                'service_results': service_results,
+                'screenshots': screenshots,
+                'backdoor': backdoor,
+                'backdoor_id': id_,
+                'comments': comments}
+    return template, args
+
+def add_new_backdoor(name, version=None, aliases=None, description=None,
+                     source=None, source_method=None, source_reference=None,
+                     campaign=None, confidence=None, analyst=None,
```
mgoffin: Should rename "analyst" to "user" which will be the NWO in newer code.

crits/backdoors/handlers.py
```diff
@@ -0,0 +1,423 @@
...
+            backdoor = Backdoor()
+            backdoor.name = name
+            backdoor.version = version
+            # Backdoor does not exist, new object. Details are handled later.
+            objs.append(backdoor)
+        else:
+            # Backdoor exists, old object. Add a new source only.
+            for s in source:
+                backdoor.add_source(s)
+            backdoor.save()
```
mgoffin: Not passing in the user to .save()

crits/backdoors/handlers.py
```diff
@@ -0,0 +1,423 @@
...
+            resp_url = reverse('crits.backdoors.views.backdoor_detail',
+                               args=[backdoor.id])
+            retVal['message'] = 'Success: <a href="%s">%s</a>' % (resp_url,
+                                                                  backdoor.name)
+            retVal['object'] = family
+            retVal['id'] = str(family.id)
+
+    # At this point we have a family object and potentially a specific object.
+    # Add the common parameters to all objects in the list and save them.
+    for backdoor in objs:
```
mgoffin: Above if we have an existing Backdoor in the database, they are not added to `objs` so any new information related to Campaigns, Tickets, Buckets, etc. will get dropped on the floor for them? What about adding a relationship to an existing family backdoor for new Backdoor entries?

crits/core/static/js/core.js
```diff
@@ -513,6 +513,7 @@ function details_copy_id (crits_type) {
         createCookie('crits_rel_id',$(this).attr('id'),60);
         createCookie('crits_rel_type',crits_type,60);
         $('button#'+readCookie('crits_rel_id')).css('background-color', '#1AC932');
+        console.log(readCookie('crits_rel_type'));
```
mgoffin: Could be removed.

crits/core/static/js/dialogs.js
```diff
@@ -1062,7 +1064,6 @@ var stdDialogs = {
       "new-target": {title: "Target", open: new_target_dialog },
 
       "backdoor_add": {title: "Backdoor"},
```
mgoffin: This is still needed?

crits/core/static/js/dialogs.js
```diff
@@ -1130,7 +1131,7 @@ var stdDialogs = {
   // action to the form's submit action by default, but I don't want
   // to make that sort of global change before 3.0.
   var singleInputDialogs = "#dialog-actor-identifier-type,#dialog-ticket,"+
-      "#dialog-backdoor_add,#dialog-source_create,#dialog-user_role,#dialog-exploit_add," +
+      "#dialog-backdoor_add,#dialog-source_create,#dialog-user_role," +
```
mgoffin: Still need backdoor here?

crits/core/user.py
```diff
@@ -89,11 +89,13 @@ class EmbeddedFavorites(EmbeddedDocument, CritsDocumentFormatter):
     """
 
     Actor = ListField(StringField())
+    Backdoor = ListField(StringField())
```
mgoffin: This was not also added to EmbeddedSubscriptions.

crits/core/user.py
```diff
@@ -89,11 +89,13 @@ class EmbeddedFavorites(EmbeddedDocument, CritsDocumentFormatter):
...
     Campaign = ListField(StringField())
     Certificate = ListField(StringField())
     Domain = ListField(StringField())
     Email = ListField(StringField())
     Event = ListField(StringField())
+    Exploit = ListField(StringField())
```
mgoffin: This was not also added to EmbeddedSubscriptions.

crits/exploits/exploit.py
```diff
@@ -0,0 +1,43 @@
...
+        "schema_doc": {
+        },
+        "jtable_opts": {
+            'details_url': 'crits.exploits.views.exploit_detail',
+            'details_url_key': 'id',
+            'default_sort': "modified DESC",
+            'searchurl': 'crits.exploits.views.exploits_listing',
+            'fields': ["name", "cve", "description", "modified", "source",
+                       "campaign", "status", "id"],
+            'jtopts_fields': ["details", "name", "description", "modified",
```
mgoffin: Would it make sense to include a column for CVE in this list?

crits/samples/migrate.py
```diff
@@ -1,9 +1,94 @@
...
+
+    if self.schema_version < 3:
+        migrate_2_to_3(self)
+
+    if self.schema_version == 3:
+        migrate_backdoors(self)
+        migrate_exploits(self)
+        self.schema_version = 4
+        self.save()
+        self.reload()
```
mgoffin: I *think* you can remove the `.save()` and `.reload()` from the previous 2_to_3 migration since it's being done here.

crits/samples/templates/samples_listing_widget.html
```diff
@@ -6,7 +6,7 @@
             <th>Size</th>
             <th>Filetype</th>
             <th>Receive</th>
-            <th>Backdoor(v)[C]</th>
+            <th>Campaign</th>
             <th>CVE</th>
```
mgoffin: Not removing CVE?

crits/settings.py
```diff
@@ -135,6 +136,7 @@
 COL_EMAIL = "email"                                       # main email collection
 COL_EVENTS = "events"                                     # main events collection
 COL_EVENT_TYPES = "event_types"                           # event types for events
+COL_EXPLOIT = "exploit"                                   # exploits
```
mgoffin: Any specific reason to make this singular vs plural?

wxsBSD: Yes. There is already a COL_EXPLOITS which is the one used by the mapreduce (at least according to the comment). I still need to go through and remove the old collections that are no longer needed before this is accepted, and add a note to UPDATING that they can be removed from the database.

crits/samples/templates/samples_listing_widget.html
wxsBSD: Removed now. Nice catch. ;)

crits/core/user.py
wxsBSD: Fixed. Thank you!

crits/core/user.py
wxsBSD: Also fixed. Thank you!

crits/exploits/exploit.py
wxsBSD: Added. Thanks!

crits/backdoors/handlers.py
wxsBSD: That is correct. If we have an existing family or specific instance backdoor we will not overwrite any of the things you mention (campaigns, tickets, buckets). My rationale for this is that I don't want to allow people to overwrite those things without explicitly acknowledging it by doing so on the backdoor details page.

You are correct that if you have an existing family backdoor and create a new specific instance backdoor that matches the family name a relationship will not be created. I will fix that now, good catch.

mgoffin: I don't want to overwrite. I want to append. For example, if an existing Backdoor has bucket "foo" and I add the Backdoor again with buckets "foo" and "bar", I would expect "bar" gets added to the existing Backdoor. Similar behavior for Campaigns and Tickets.

wxsBSD: OK, I think that is reasonable. It will probably let me simply that handler a bit too. Will update soon.

wxsBSD:
The corresponding PR for services is at https://github.com/crits/crits_services/pull/120.

crits/samples/migrate.py
wxsBSD: OK, removed. Good catch. :smiley:

crits/backdoors/handlers.py
wxsBSD: Updated to reflect the discussion here. It should now behave as you expect, which I think is a much better way too so thank you!

crits/backdoors/handlers.py
wxsBSD: Done.

mgoffin:
When I migrated, I had a sample with a Backdoor "Foo v1.0" and an Exploit "CVE-FOO". After the migration I had the following:

- Sample had backdoor and exploit removed (expected)
- Backdoor Foo created (expected)
- Backdoor Foo v1.0 created (expected)
- Backdoor Foo and Foo v1.0 related to each other (expected)
- Exploit CVE-FOO created
- Sample related to Backdoor Foo v1.0 (expected)
- Sample related to Exploit CVE-FOO (expected)

I did not, however, get Sample related to Backdoor Foo, which was expected.

wxsBSD:
I believe I've addressed your concern. Now when relating anything to a versioned backdoor we will attempt to find a family backdoor and also relate to that.

crits/backdoors/handlers.py
```diff
@@ -0,0 +1,449 @@
...
+        source = [create_embedded_source(source,
+                                         reference=source_reference,
+                                         method=source_method,
+                                         analyst=user)]
+    elif isinstance(source, EmbeddedSource):
+        source = [source]
+
+    if not source:
+        retVal['message'] = "Missing source information."
+        return retval
```
mgoffin: Should be retVal. Also a couple unused imports here. Might need to cleanup imports all over if they were copied from other files.

jezkerwin:
I updated this morning and followed the steps within UPGRADING. I udpated the STIX and CyBox python packages, restarted apache and Mongo. The TLOs appear in the dashboard, however when I click 'Add Backdoor' or 'Add Exploit', nothing happens.
I've gone into other TLOs and those still work.

Any thoughts on what might be happening?

wxsBSD:
Do you see any requests being made? It should fetch the dialog.

jezkerwin:
FireBug shows no requests being made, the CSV and Reload options works, just not the Add Exploit or Backdoor.

mgoffin:
Clear your cache :)

jezkerwin:
Yep, did that. No good.

jezkerwin:
I did see this error in the ssl_error_log, do you think it could be related?
---
mod_wsgi (pid=8371): Exception occurred processing WSGI script '/data/crits/django.wsgi'., referer: https://<fqdn>.local/backdoors/list/
IOError: failed to write data, referer: https://<fqdn>.local/backdoors/list/
---

mgoffin:
Most of the stuff I'm seeing online are not related to code but to client/server interaction. There's no clear answer but I'm not 100% certain its the code. You'd have to do some digging on that as I can't replicate at the moment.

jezkerwin:
Thanks, I'll do a bit more digging on my end and see what I can find.

Issue #442, opened 2015-04-02
String Indicator does not properly export to STIX


brlogan:
When exporting a Indicator of type String to a STIX document, the resulting XML does not include the Indicator value.

brlogan:
After some poking, I'm still not certain how to correctly handle this, so I'm going to bug @gtback again. :)

gtback:
@brlogan - Is this also addressed by #493? Is there anything else you need from me regarding this one?

brlogan:
Thanks for the reminder, Greg.
I don't need anything else from you at the moment, but I will be sure to let you know if I think of something.

Issue #437, opened 2015-03-31
Email Attachments not included during STIX Import/Export


brlogan:
When an Email is exported from CRITs to a STIX document, the email's attachments are not included. Similarly, if a STIX document contains an Email with attachments, those attachments are not imported into CRITs.

brlogan:
Due to a bug in python-cybox, relationships are not properly maintained when using the `add_related` function to generate a STIX document. See CybOXProject/python-cybox#237 for further details.

mgoffin:
CybOX/STIX are no longer a part of core. They are a part of the TAXII service and if there are issues now that the code has moved, new ones should be created in the `crits_services` Issue tracker.

PR #493, opened 2015-05-07
Sharing String Indicators and Relationships via STIX

brlogan:
This PR is an attempt at enabling the sharing of String Indicators and TLO relationships via STIX.

Strings:
- There is no CybOX object type designed for generic strings
- This solution uses the CybOX Custom object
  - Object custom_name = "crits:String"
  - The single custom_property name is "value"
  - This requires an update to cybox-2.1.0.11 and stix-1.1.1.5
  - This is intended to address #442

Relationships:
- In order to describe via STIX the inter-object relationships, all Observables are wrapped within a STIX Indicator (I'm not a huge fan of this Observable wrapping and I expect @mgoffin won't be either, so I'm open to suggestions :smiley:)
  - The STIX Indicator title and description indicate that the Indicator is simply a wrapper
  - The STIX Indicator's related_indicators attribute describes the relationships using references to other STIX Indicators
  - We could only wrap those Observables that actually have relationships, but I thought it might be better to consistently wrap all of them
- Email attachments are automatically included with an Email
  - The Sample is added as its own Indicator-wrapped Observable
  - The Sample is referenced in the "attachments" attribute of the Email
  - The relationship between the Email and the Sample is documented in the related_indicators attribute of their respective Indicator wrapper
  - This attempts to address #437 
- In order to describe via STIX the Event -> Object relationships, an Incident is added to the STIX output. (I'm not a huge fan of this either, so again, I'm open to suggestions)
  - Relationships to Indicators and Indicator-wrapped Observables are listed in the Incident's related_indicators attribute.
  - Any objects for which a relationship to the Event is not defined are assigned a "Related_To" relationship.

### DIFF
--- a/UPDATING
+++ b/UPDATING
@@ -3,6 +3,17 @@ should be aware of. This will also include things like dependency changes or
 references to changes in other repositories. Updates are sorted by commit date
 descending.
 
+Author: brlogan
+Date: 2015-05-07
+
+    Added support for sharing String Indicators via STIX. Because there is no
+    generic String type in CybOX, the CybOX Custom object must be used. This
+    requries updating python-stix and python-cybox.
+    
+    Updated versions are:
+        cybox-2.1.0.11
+        stix-1.1.1.5
+
 Author: mgoffin
 Date:   2015-04-28
 
--- a/crits/core/crits_mongoengine.py
+++ b/crits/core/crits_mongoengine.py
@@ -698,6 +698,7 @@ class CritsDocument(BaseDocument):
 
         # Store message
         stix_msg = {
+                       'stix_incidents': [],
                        'stix_indicators': [],
                        'stix_observables': [],
                        'stix_actors': [],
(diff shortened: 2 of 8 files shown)


mgoffin:
Partial aside to this, but the custom "crits:Foo" object...how hard was it to implement? As in, if it was simplistic to replicate can we "backfill" support for the rest of the indicator types into STIX/CybOX that we don't have great support for? Would also pave the way for supporting a larger vocabulary instead of waiting for CybOX to implement things?

brlogan:
It's very simple to implement. You instantiate the Custom object, set a name and description, then add any custom properties you need. For the case of Indicators, the only property would probably be "value".
It shouldn't take too much work to build out any types that aren't currently supported by python-cybox.

mgoffin:
So in the future if we wanted to expand on Indicators (add fields, alter vocabulary, etc.) we can change the object_mapper to be amazingly simple (at least in the CRITs -> CybOX direction):

- Create custom object
- Jam in attributes

And we could do this for every single Indicator type CRITs supports? What sort of custom code would be required to know how to parse out our custom object on the other end (if someone wasn't using CRITs to parse the STIX document)?

brlogan:
It probably makes sense to use the standard types in python-cybox wherever possible for optimal interoperability, but I don't see why we couldn't use the Custom object for everything else. It would be very useful for CRITs <-> CRITs sharing while still being valid CybOX. Indicator types whose values aren't just strings may be somewhat more complex, but there aren't many of those.

As for parsing the Custom object from STIX, it would of course depend on the use case, but they could grab any number of properties with something like this:
```python
if isinstance(cybox_obj, Custom):
    properties = {}
    object_name = cybox_obj.custom_name
    for prop in cybox_obj.custom_properties:
        properties[prop.name] = prop.value
```

mgoffin:
Cool. So it would be 10000x easier to do CRITs->CRITs STIX sharing, but when it comes to consuming STIX generated by other sources, all bets are off.

brlogan:
Pretty much.
I don't think it would be a good idea to solely use the Custom object for everything, but I think it could be a good solution for those types that aren't currently supported in python-cybox. CRITs instances would be able to consume the Custom objects correctly, and because the Custom objects are valid CybOX, the worst that *should* happen with other STIX-consuming systems is that those objects just get dropped.

mgoffin:
Which version of CybOX does this require? In trying to use this PR I am getting a traceback because:

```python
    from cybox.objects.custom_object import Custom
ImportError: No module named custom_object
```

If it requires a new version we should put that in requirements.txt as well as make an UPDATING entry.

gtback:
The custom object was added in 2.1.0.10: 
https://github.com/CybOXProject/python-cybox/releases/tag/v2.1.0.10

The latest version is 2.1.0.11

mgoffin:
Yea then if that and whatever STIX version is required both work with CRITs then UPDATING should get an entry to install the right versions and requirements.txt should get the changes as well.

brlogan:
Darn it, I can't believe I forgot to even mention that. Both STIX and TAXII should be updated for this to work. I won't have access to my DEV laptop until Monday night, but I'll see if I can add in those updates before then.

mgoffin:
I tried this out. Got this with an Indicator of type "String":

```
Imported
Event	554d4f8303b6e0145efdb0e2
Failed
Custom	ID: example:Custom-ddb9cb3d-76f1-426c-a8b7-0513ae4e1ec5

Can't create indicator with an empty type field
```

This is pulled from the STIX doc I tried to upload:

```xml
    <stix:Indicators>
        <stix:Indicator id="example:indicator-e2d11d22-5698-4154-8fd0-ce4d7edc2cdd" timestamp="2015-05-09T00:06:05.269136+00:00" xsi:type='indicator:IndicatorType'>
            <indicator:Observable id="example:Observable-bb34dd7e-b48c-402f-b1f8-a891b1fff78e">
                <cybox:Object id="example:Custom-ddb9cb3d-76f1-426c-a8b7-0513ae4e1ec5">
                    <cybox:Properties xsi:type="CustomObj:CustomObjectType" custom_name="crits:String">
                        <cyboxCommon:Custom_Properties>
                            <cyboxCommon:Property name="value" description="Generic String">this is an indicator</cyboxCommon:Property>
                        </cyboxCommon:Custom_Properties>
                        <CustomObj:Description>This is a generic string used as the value of an Indicator or Object within CRITs.</CustomObj:Description>
                    </cybox:Properties>
                </cybox:Object>
            </indicator:Observable>
            <indicator:Producer>
                <stixCommon:Identity>
                    <stixCommon:Name>Home</stixCommon:Name>
                </stixCommon:Identity>
            </indicator:Producer>
        </stix:Indicator>
    </stix:Indicators>
```

mgoffin:
@brlogan Any ideas on the above problem?

brlogan:
Sorry for the delay and the trouble. I was out Thursday-Monday.

It looks like I missed a couple function calls when merging Master. It should work now.

PR #495, opened 2015-05-12
Bootstrap fixes for OSX

frbapolkosnik:
...
 from crits.core.mongo_tools import mongo_connector
  File "/Users/adam/Downloads/crits/crits/core/mongo_tools.py", line 8, in <module>
    import magic
  File "/Library/Python/2.7/site-packages/magic.py", line 161, in <module>
    raise ImportError('failed to find libmagic.  Check your installation')

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -67,7 +67,7 @@ then
     echo "Installation for OSX requires Homebrew. Please visit http://brew.sh/."
     exit
   }
-  brew install mongodb ssdeep upx p7zip libyaml pcre libxml2 openssl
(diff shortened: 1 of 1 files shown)


frbapolkosnik:
I was told that libmagic should be installed by default, so I checked.
I tried this on a fresh install of Yosemite (straight from the recovery image) and still libmagic was not there.

Additionally, I think that there should be a check for presence of pip, in the same way that presence of brew is checked, but perhaps we can run "sudo easy_install pip" to install pip.

If pip (and the python packages are not installed as a result) is not installed, the database.py will end up with empty SECRET_KEY.

PR #499, opened 2015-05-15
Sector backdoor and exploit

wxsBSD:
Add backdoor and templates to sector jtable.

Add backdoor and exploit counts to sector objects (default to 0).

Add sector to backdoor and exploit templates.

### DIFF
--- a/crits/core/sector.py
+++ b/crits/core/sector.py
@@ -29,7 +29,6 @@ class Sector(CritsDocument, CritsSchemaDocument, Document):
             'Event': 'Integer',
             'Exploit': 'Integer',
             'Target': 'Integer',
-            'Event': 'Integer',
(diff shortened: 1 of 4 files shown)


crits/core/sector.py
```diff
@@ -21,10 +21,13 @@ class Sector(CritsDocument, CritsSchemaDocument, Document):
...
             'Actor': 'Integer',
+            'Backdoor': 'Integer',
             'Campaign': 'Integer',
             'Certificate': 'Integer',
             'Domain': 'Integer',
             'Email': 'Integer',
+            'Event': 'Integer',
+            'Exploit': 'Integer',
             'Target': 'Integer',
             'Event': 'Integer',
```
mgoffin: It's here twice?

PR #506, opened 2015-05-20
Bootstrap updates

frbapolkosnik:
Summary of changes:
- addirng Universe repo for Ubuntu. Without universe repo some of the dependencies are missing on Ubuntu.
- Generate the random key without using the python script. When apt-get would fail, the user would end up with an empty SECRET_KEY, and more failures that are harder to troubleshoot. 
- Change unrar to unrar-free to fix ubuntu-15.04
- remove an extra space

We can now probably retire crits/contrib/gen_secret_key.py

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -108,7 +108,7 @@ chmod 664 logs/crits.log
 if [ ! -f crits/config/database.py ]; then
   echo "Creating default database configuration file"
   cp crits/config/database_example.py crits/config/database.py
-  SC="$(cat /dev/urandom | tr -dc 'abcdefghijklmnopqrstuvwxyz0123456789!@#%^&*(-_=+)' | fold -w 50 | head -n 1)"
+  SC=$(cat /dev/urandom | tr -dc 'abcdefghijklmnopqrstuvwxyz0123456789!@#%^&*(-_=+)' | fold -w 50 | head -n 1)
   # This is going to escape the '&' character that is a special character in sed 
   SE=$(echo ${SC} | sed -e 's/&/\\\&/g')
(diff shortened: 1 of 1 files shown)


mgoffin:
I get an "Illegal byte sequence" with `tr` on OSX.

I suspect we can keep the "gen_secret" script as it can be useful in other applications, or if someone needs to re-roll their secret and doesn't want to have to rerun bootstrap.

apolkosnik:
Good catch!
I am sure glad that Solaris is not in the scope ;-)

I'll fix it a little later today, as I'm nowhere near my OSX test boxes.
Thanks!

I get an "Illegal byte sequence" with tr on OSX.

I suspect we can keep the "gen_secret" script as it can be useful in other
applications, or if someone needs to re-roll their secret and doesn't want
to have to rerun bootstrap.

—
Reply to this email directly or view it on GitHub
<https://github.com/crits/crits/pull/506#issuecomment-103861374>.

apolkosnik:
I think that prepending tr with LC_CTYPE=C should do the trick.
On May 20, 2015 8:20 AM, "Mike Goffin" <notifications@github.com> wrote:

I get an "Illegal byte sequence" with tr on OSX.

I suspect we can keep the "gen_secret" script as it can be useful in other
applications, or if someone needs to re-roll their secret and doesn't want
to have to rerun bootstrap.

—
Reply to this email directly or view it on GitHub
<https://github.com/crits/crits/pull/506#issuecomment-103861374>.

frbapolkosnik:
ok, tested the fix for OSX, and it didn't break it on Ubuntu.

PR #507, opened 2015-05-21
Replicaset option

ManoftheSea:
When connecting to a replicaset, the connection needs a single name.  If that host is not the Primary, things die horribly.  By adding the replicaset option, the system can be configured to auto-discover the other nodes, and connect to the primary (due to ReadPreference=PRIMARY)


ManoftheSea:
Sorry, I thought there was a way to cut it down to just the replicaset option, without the merge from crits/master... but I guess it doesn't matter.

mgoffin:
This doesn't expose the options in the database_example.py file, where people should be overriding if they need something other than the default.

How does this work when you pass in a 0-length string as the replicaset?

ManoftheSea:
Added the option to the database_example.py file, with the default setting None.
Changed the option in settings to None from empty string.

When set to None, the MongoClient connection treats it as no replicaset.  When configured with a name, the "hosts" are treated as seeds, the other entries are discovered, and reads/writes are directed to the primary node. (In accordance with existing ReadPreference=PRIMARY)

ManoftheSea:
Oh hey, it turns out you can also turn on replicaset behaviour just by passing a list of hosts, instead of a single one.
But this allows use of a replicaset with a single seed.  Also, not a big change.

PR #510, opened 2015-06-03
Bootstrap for CentOS 6.4, 6.5 and 6.6

pixel5919:
Will install and start an instance of CRITs. Only some services will be
installed.


mgoffin:
Thanks for contributing!

A couple things I want to bring up:

- It would be much more beneficial if this was integrated into the bootstrap script that exists already for handling installation. It's much easier for people to follow the single install step than to have to determine which script to use.
- I would refrain from installing *any* services. Services are optional and we shouldn't force people into installing them.

pixel5919:
Thanks for the feedback! I'll get to work on it right now.

pixel5919:
I am finishing up writing the script and I have several questions:

In the original bootstrap, there is a prompt for the user to continue in starting MongoDB and the Server, do we want that included?

For CentOS, there is a lot of installation so when I wrote the original, I just did sudo yum -y install to avoid having the user sit for an hour waiting to type in y in order for CRITs dependencies to install. Should I leave it that way for at least the CentOS installation or try to keep the installation consistent with the other Linux Distributions?

frbapolkosnik:
This is pretty nice. I like the modular approach there. The current script runs like the russian ice breaker ;-)

mgoffin:
A recent PR included all of this to make CentOS work

PR #511, opened 2015-06-04
Fix dependencies for CentOS 7

RichieB2B:
Two additionally repositories need to be added before this bootstrap will work. Should the bootstrap check/fix this? Or should this be noted as a prerequisite in the documentation?

http://download.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm
http://pkgs.repoforge.org/rpmforge-release/rpmforge-release-0.5.3-1.el7.rf.x86_64.rpm

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -26,7 +26,7 @@ if [ -f /etc/lsb-release ]; then
   OS=$DISTRIB_ID
   VER=$DISTRIB_RELEASE
 elif [ -f /etc/redhat-release ]; then
-  OS=$(cat /etc/redhat-release | sed 's/ Enterprise.*//')
+  OS=$(cat /etc/redhat-release | sed 's/ Enterprise.*//;s/ Linux.*//')
   VER=$(cat /etc/redhat-release | sed 's/.*release //;s/ .*$//')
 else
   OS=$(uname -s)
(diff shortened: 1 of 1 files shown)


mgoffin:
There's a lot of things you are now installing using Yum which should be handled using pip later on.

frbapolkosnik:
RichieB2B,
Thanks for contributing!

You should let the following to be installed through pip ( as that way we can at least make sure that appropriate versions of those packages are installed, like mongoengine..., and it's much easier to maintain the sane install/upgrades with python packages managed through pip):

python-lxml python-mongoengine python-dateutil python-ldap python-pyparsing python-magic python-mimeparse pytz PyYAML python-requests python-simplejson

I've gone through that painful process on Ubuntu where I was trying to install as much as possible from apt-get, and in the end I ended up with some packages that were not up the snuff. Take a look at what is installed on Ubuntu, and try to mirror that, let the pip install the python-* packages.

Best regards,
Adam

frbapolkosnik:
You can add/enable these repos like this:

`yum-config-manager --add-repo repository_url`

`yum-config-manager --enable repository…`

RichieB2B:
I prefer using distribution packages where possible. If specific or minimum versions of modules are required those version numbers should be included in requirements.txt (which they are). If the system packages do not satisfy the version requirements pip will pull them in, but if in the future the right version appear in a distribution those will be preferred instead. 

@frbapolkosnik mentions upgrades are easier to handle with pip, does this mean https://github.com/pypa/pip/issues/59 is fixed?

Also, yum-config-manager --add-repo requires the URL of a repo file. Most repositories distribute those in an rpm instead.

Either way, I just wanted to contribute a working bootstrap for CentOS 7 which includes python 2.7. This seemed a lot easier than getting python 2.7 to work on CentOS 6.

pixel5919:
Python 2.7 can work for CentOS, you just have to install it like this:

sudo yum -y update
    sudo yum -y groupinstall "Development tools"
    sudo yum -y install zlib-devel bzip2-devel openssl-devel ncurses-devel sqlite-devel readline-devel tk-devel gdbm-devel db4-devel libpcap-devel xz-devel php-ldap
    sudo yum -y install xz-libs

 wget http://www.python.org/ftp/python/2.7.9/Python-2.7.9.tgz
        tar -xzf Python-2.7.9.tgz

        cd Python-2.7.9
        ./configure --prefix=/usr/local --enable-unicode=ucs4 --enable-shared LDFLAGS="-Wl,-rpath /usr/local/lib"
        sudo make && sudo make altinstall
        cd ..
        sudo rm -rf Python-2.7.9.tgz
        sudo rm -rf Python-2.7.9
        # Add Python To Path
        cd /usr/local/bin
        sudo ln -s python2 python
        sudo ln -s python2.7 python2

RichieB2B:
CentOS 7 comes with python 2.7 by default. No need to compile it yourself.

pixel5919:
Sorry, I meant that is how you do it for CentOS 6.

mgoffin:
A recent PR has combined all of the things to make CentOS work

PR #519, opened 2015-06-18
Support STIX documents with inline embedding

RichieB2B:
I used a STIX export from https://github.com/MISP/MISP to develop and test this patch.

This is work in progress and needs testing to see if a fully related STIX documents still imports ok.

Also, the use of parse_observables() vs make_crits_object() needs to be more intelligent. Now both are done which can't be good.

### DIFF
--- a/crits/standards/parsers.py
+++ b/crits/standards/parsers.py
@@ -149,21 +149,43 @@ class STIXParser():
                 if self.package.incidents:
                     incdnts = self.package.incidents
                     for rel in getattr(incdnts[0], 'related_indicators', ()):
-                        self.event_rels[rel.item.idref] = (rel.relationship.value,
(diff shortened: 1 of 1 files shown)


mgoffin:
Based on the discussions on the development list and the active development work going on, I am hesitant to accept any code related to CybOX/STIX as all of the code is being ripped out of core and being moved into the TAXII service.

RichieB2B:
As long as the new code will support inline STIX documents that is fine with me. ;-)

I needed this patch to connect MISP to CRITS.

mgoffin:
No clue. I don't plan on making the code any better. I'll leave that up to people who plan on using it. My goal is simply separating.

RichieB2B:
I'll submit a new patch if it is not the case (and we're still using MISP and CRITS together).

PR #526, opened 2015-07-02
Vis.js relationship grapher from Nate (github: Magicked)

ckane:
Nate integrated vis.js into the branch they are using for CRITs at his company. I pulled these into master and have been playing with it on my personal instance, tracking master, and felt it is already at a good point to send over to you all to consider for incorporation into the main project.

screenshot from 2015-07-02 08 20 14


mgoffin:
This looks neat! What's the difference between vis.js and the already-in-CRITs D3 JS library?

mgoffin:
And how does this differ from the Relationships service tab? :)

ckane:
I have always struggled with using D3 to build this type of visualization, and other peers have expressed similar issues. Vis.js ended up seeming to be a cleaner, if less feature-rich, option. I don't know how to access either of the two functions you mentioned above in crits master... can you point me at them and I can probably write a semi-detailed description of what I think to be different.

mgoffin:
As long as you are using services and you have a somewhat recent version of crits_services installed, it comes with the Relationships service (https://github.com/crits/crits_services/tree/master/relationships_service). It adds a tab to each TLO which provides a graph similar to what you have pictured. I haven't looked at this in a UI yet to know specifically what's different. I do know that the Relationships service has been worked on for a while and it was recently updated to allow you to attribute a graph to a Campaign.

mgoffin:
Went to go try this out and it breaks everything because I can't log in. Looks like an accidental inclusion of a change to the login URL in settings.py.

Magicked:
Oh yea, sorry about that. That's something I haven't fixed yet with our instance.

ckane:
Oh I can try pulling that out...

ckane:
How's that?

ckane:
I think one of the chief differences which Nate applied here (in contrast to relationship explorer) is that this visualization was incorporated as a top-level feature accessible directly from an object details view. Additionally, vis.js offers really nice built-in, optional, graph editing mechanisms which we are investigating as an avenue to provide "visualize->manually filter/exclude->apply bulk action" workflows around. The javascript to retrieve "current graph state" is pretty clean from the handful of test-cases I've been working with here. I'm not sure what/if there's an analog to this in D3.

mgoffin:
So both are accessible on a TLO's Details page. The Relationships graph allows you to filter by TLO Type as well as set a max traversing depth, and bundle all of the graph into Campaigns. Another benefit to it being a service and not in core is that it can be enhanced and made available without having to release a new version of CRITs (to those people who track releases).

Just playing devil's advocate here to see what the Relationships Service didn't provide that required building what seems to be a replication of the feature but put into core and with a different JS library. Not to take away from the great work because the code looks very clean and easy-to-maintain.

ckane:
That's totally fair... I'm installing all the stuff from crits/crits_services right now on my local master and will provide more feedback.

ckane:
After playing with it for a bit, I think that you are correct from the "current state" that the functionality is superseded by the existing relationship_service. The built-in editor within vis.js adds a nice author-supported feature which I can only find implemented by random other authors for d3. Additionally, IMO, the code and the graph rendering using vis.js is more aesthetically pleasing.

What would be your opinion of an attempt to replace the d3 code in relationships_service with a functional equivalent built from vis.js?

mgoffin:
I wouldn't be opposed to it :) I know I used D3 to create the Roles graph (in core) that's only in the Roles branch currently. However, we could have both libraries for now until I can either port it to vis.js or give up and we keep both libraries around.

ckane:
Having alternatives isn't always a bad thing. I know that D3 offers a LOT more freedom in presentation options, and that's reflected in the verbosity of the diagram set-up code. That said, Vis.js appears to offer some other out-of-the box features that we're looking into for enabling data manipulation work, without all of the setup overhead of D3, since it seems to default to some decent "good enough" initial values for rendering.

I'll play with it a bit and probably kick a PR over to the other project in that case.

mgoffin:
Sounds good! I'd recommend adding vis.js to the core project in the same place D3 lives and leverage it in the service. That way if other people want to use it they can use the one from core and not need to include it themselves.

ckane:
Sounds good, will do

ckane:
Ok I reset my master and then turned this into a PR just to pull the vis.js code into the main project

packet-rat:
Coleman/Mike,

Many thanks.  Good stuff.

Patrick Maroney
Cell: (609)841-5104


On Jul 3, 2015, at 1:14 PM, Coleman Kane <notifications@github.com> wrote:

Ok I reset my master and then turned this into a PR just to pull the vis.js code into the main project

ckane:
If you'd like to follow along:
* https://github.com/Security513/crits_services/tree/visjs_relationship_service

ckane:
first-rev-visjs

Issue #481, opened 2015-05-04
WebService API Guidelines for Sources.


cdorer:
I'm working on a need to create a WebService API that will provide a listing of all Sources based on the user.  This is intended for helping with data validation when using pycrits in creating TLOs, Samples for instance.

There are the two facts I'm working with.

* Copy and pasting the code from https://github.com/crits/crits/blob/master/crits/samples/forms.py#L89 to line 92 into the Sources/api.py class I'm building.
    * I don't see an abstraction call in an another place so copy and pasting is bad form; Is there another place I could put the code where it can be referenced in the future by other modules?
* Utilizing crits.core.api CRITsAPIResource.crits_response and providing a list of the source options.
    * What's the best practice of structuring the data structure for this list in the CRITs way?

PR #529, opened 2015-07-06
Source api call

cdorer:
Provides the list of sources available to the user
that's in the system.

Great for helping provide validation of the
creation of TLO's through the webservice api

refer to #481


cdorer:
I closed the last one because of branching I'm doing with my repo.  Restarting a fresh instance.  Anything I can do to help, let me know.

mgoffin:
Could you describe a bit more what use cases you have for this? I think acquiring "what do i have access to?" is good, but I think it might be better go to about this a different way.

My thought is why stop at sources? What if we allowed to query for all of the stuff you have access to and can do? Basically an API call for your profile?

cdorer:
This particular use case, I only needed a list for the available sources for that user, which provided the purpose for the cli tool.  We have lazy users and I thought the best way was to check, at an atomic level, which sources available and nothing else.

mgoffin:
I think from a long-term use and maintenance perspective there's a few things:

- having a new "sources" folder which only handles this small use-case will be confusing for managing and maintaining code for sources
- it would be much more useful for a user to query their profile and get back a lot of content, for example:
  - sources they have access to
  - things they subscribed to
  - favorites
  - notifications

cdorer:
I can dig that. It can go either way, and you're suggesting a more logical approach if the sources concept stays small.

It definitely can be more resource intensive if the call goes for the whole kitchen sink, instead of a portion of what the user wants.

mgoffin:
Not really. If you only want sources, then limit the query:

https://github.com/crits/crits/wiki/Authenticated-API#limiting-get-request-results

cdorer:
Sweet, that works.

packet-rat:
[+1]

Patrick Maroney
(609)841-5104

On Jul 31, 2015, at 2:37 PM, Mike Goffin <notifications@github.com> wrote:

I think from a long-term use and maintenance perspective there's a few things:

having a new "sources" folder which only handles this small use-case will be confusing for managing and maintaining code for sources
it would be much more useful for a user to query their profile and get back a lot of content, for example:
sources they have access to
things they subscribed to
favorites
notifications

Issue #530, opened 2015-07-06
Memory Hog When Deleting Indicators


cdorer:
Found when indicators are deleted it takes time and eats up a ton of memory.  The big bottleneck was when Objects of the Indicators were removed it brute forced file look ups especially when they weren't of enum type 'file'. mongoengine looked up and cached all files of the TLO's to ensure there were no other references before deleting the file.

In cases of LARGE databases this took minutes to delete one indicator and ate up gigs of memory.

Happens if query caching is on or off!

When deletes are happening in bulk with web browsers by smart users, this has the capability to knock out a server QUICKLY.

PR #531, opened 2015-07-06
Removed a huge memory issue

cdorer:
Do file checks when objects are of file datatype and
ensure filenames are of md5sum hashes when looking
to delete files.

Found when indicators are deleted it takes time and eats up
a ton of memory.  The big bottleneck was when Objects of the
Indicators were removed it brute forced file look ups
especially when they weren't of enum type 'file'.
mongoengine looked up and cached all files of the TLO's to ensure
there were no other references before deleting the file.

In cases of LARGE databases this took minutes to delete one
indicator and ate up gigs of memory.

https://github.com/crits/crits/issues/530

### DIFF
--- a/crits/core/crits_mongoengine.py
+++ b/crits/core/crits_mongoengine.py
@@ -1608,7 +1608,8 @@ class CritsBaseAttributes(CritsDocument, CritsBaseDocument,
 
         from crits.objects.handlers import delete_object_file
         for o in self.obj:
-            delete_object_file(o.value)
+            if o.datatype == 'file':
+                delete_object_file(o.value)
         self.obj = []
 
     def delete_all_favorites(self):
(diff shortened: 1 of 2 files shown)


mgoffin:
Since this entire section of code is redone in an existing PR that we are looking to merge, I'll make this change in that PR. Keep this open but don't merge it as a reminder to me when I get back to make this change.

Thanks!

mgoffin:
Got your change added to the PR! I'll close this out. Thank you for bringing it up and getting the solution in place!

cdorer:
Thank you! Which PR is the one you're working on?

mgoffin:
`remove_cybox_stix` for core, and `add_cybox_stix` for crits_services (if you use services you have to have this branch otherwise things will break).

PR #534, opened 2015-07-08
Update bootstrap's SECRET_KEY handling for OSX #2

frbapolkosnik:
Now we fix a "double backslash" case.

I've kicked off a test with couple of 100k iterations and there were no more errors.


gtback:
Would it be easier to just remove '&' from the list of characters in the SECRET_KEY alphabet a couple lines higher up? :wink:

apolkosnik:
I think that the last commit yesterday did the trick.

PR #540, opened 2015-07-20
Add support for a configurable ZIP password for downloaded samples

frbapolkosnik:
This will allow to change the password for zip files that CRITs creates to be set in the global configuration.


crits/config/forms.py
```diff
@@ -71,6 +71,10 @@ class ConfigGeneralForm(forms.Form):
         widget=forms.TextInput,
         help_text='*Requires a web server restart.',
         required=True)
+    zip7_password = forms.CharField(
+        widget=forms.TextInput,
+        help_text='*Requires a web server restart.',
```
mgoffin: This shouldn't require a web restart.

mgoffin:
I think this looks good. It doesn't need a web restart to take effect since you are querying the config live every time a zip is made. I'd remove that help text and we should be good to accept!

frbapolkosnik:
ok, I removed the help textl, and added descriptive label for the setting

Issue #491, opened 2015-05-06
Uploading zip or rar via the API fails


wxsBSD:
Uploading a zip or a rar file via the API results in a failure. Here's an example:

```
wxs@psh tmp % curl -F "filedata=@foo.zip" -F "source=foo" -F "file_format=zip" -F "upload_type=file" "http://localhost:8000/api/v1/samples/?username=wxs&api_key=<key>"
```

```
File "/Users/wxs/Documents/dev/crits/crits/samples/api.py", line 114, in obj_create
    if result.get('message'):

AttributeError: 'str' object has no attribute 'get'
```

Lambdanaut:
Is this fixed in master yet? If not I may be tackling it tomorrow.

Lambdanaut:
Fixed in https://github.com/crits/crits/pull/543

Please allow the pull request :D

PR #543, opened 2015-07-27
Fixes issue #491 so the sample api doesn't bug out

Lambdanaut:
This is admittedly a very small change. The samples api looks to be unfinished. There are
a few lines of code referencing keys in the results dictionary that just can't possibly exist.

Anyways, this fixes the bug. :D In the future the samples api needs to be cleaned up.


mgoffin:
To be honest, the right fix (as yours still can be broken) is to not use `handle_uploaded_file` if all of the potential return types aren't going to be explored, and instead use `handle_file`.

mgoffin:
I should probably add that my personal preference is to handle all of the return types on `handle_uploaded_file` so we can keep the explosion of rar/zip support :)

Lambdanaut:
Okay. So should I make that fix, or will this be merged? Wondering where to go from here.

Lambdanaut:
I see that handle_uploaded_file will return a list if the type is zip or rar, and a dictionary otherwise. That's where the error is coming in. I'll make a commit to deal with either of these cases.

mgoffin:
It still returns a list:

https://github.com/crits/crits/blob/master/crits/samples/handlers.py#L1099

Lambdanaut:
Correct. Misread that.

Lambdanaut:
Thanks Mike, I was looking at the different returned objects from handle_file, handle_zip, handle_rar etc etc etc. 

What would be ideal is if the JSON that the request returned included the uploaded object's ID for every object uploaded, whether it was a zip, rar, or whatever.

If I did that, it seems like it'd need to return a list of IDs instead of just one. What should I call that field in the returned JSON?

 Can I make that change and get it merged?

mgoffin:
I think what handle_file returns is what's jammed into the list, which iirc should be a dictionary with all of the stuff you're looking for (i am not in front of code right now).

Lambdanaut:
Mike. I added a commit to this pull request to let it handle both cases. Could you check it out and let me know if I should make any more changes? 

Thanks :)

mgoffin:
Yep! I'll try to get to this in the morning and let you know :)

Lambdanaut:
Get any chance to look at it?

crits/samples/api.py
```diff
@@ -113,21 +113,39 @@ def obj_create(self, bundle, **kwargs):
...
+                if result.get('success'):
+                    content['return_code'] = 0
+
+            else:
+                # Metadata for Zip, RAR, etc
+                content['id'] = result
+                page_url = reverse('crits.samples.views.detail',
+                                   args=[result])
+
+                content['page_url'] = page_url
```
mgoffin: What is this for?

Lambdanaut: Apologies for over-verbosty: 

Line 117 is going to get the first item in the list of results returned by `handle_uploaded_file` 

The first item will either be a string that is the ID of the uploaded sample or an object representing the upload with loads of data about it. 

One side of the if-then-else statement deals with the object. The other side deals with a string of the ID. 

I want to get a link to the sample uploaded to the API so `content['page_url']` provides that. Is this a good way to go about that?

mgoffin: Why not just use `url` like everything else though? Having this random extra `page_url` field that isn't common for return values across all TLOs seems odd to me so I was trying to figure out why it was needed :)

Lambdanaut: because `url` is an API endpoint we can hit to get data about the sample but we have an application that will need to somehow get a direct link to the sample's Crits page or the download link for the sample itself.

mgoffin: If it's unique to an application that you have then it definitely shouldn't belong in the code base.

Lambdanaut: Gotcha. Didn't quite grok the philosophy for the Api. So it shouldn't return page_url. I'll rip that out.

mgoffin: It should still return a `url`, which should be the same type of url as it returns in other cases (the one you removed).

Lambdanaut: Right :+1:

Lambdanaut:
Made a commit removing page_url. I had to get the id of the sample object to return over the request, so I hit Mongo for that. Could you give it a lookover?

Lambdanaut:
Made a final commit 0b53935 reverting everything else and applying this fix in a simpler way. I think this is The Right Way. It's all in the commit message. Has been tested on raw, zipped, and rar files. Let me know.

brlogan:
Hey @mgoffin, are you okay with this change? I want to make sure it's going to be pulled in before I use it.

mgoffin:
I haven't been able to test it yet. There are situations where we return lists of samples added to the system and I don't want to break those because we are returning a single MD5 instead of a list.

PR #547, opened 2015-07-29
Fix get_dialog with invalid dialog.

wxsBSD:
Don't throw a 500 if someone asks for an invalid dialog. Instead return
the error page.

### DIFF
--- a/crits/core/views.py
+++ b/crits/core/views.py
@@ -185,9 +185,16 @@ def get_dialog(request):
 
     dialog = request.GET.get('dialog', '')
     # Regex in urls.py doesn't seem to be working, should sanity check dialog
-    return render_to_response(dialog + ".html",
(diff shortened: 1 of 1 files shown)


mgoffin:
The dialogs are acquired via AJAX. How does returning the error page help?

mgoffin:
Seems to me that if the dialog templates being returned were included as-is on the form, returning the error.html page would include the entire CRITs UI back to be included in the form? I'd think we would instead want to return some useful text that can be included in the form, but I also don't know what you originally saw to trigger this fix.

wxsBSD:
Good point. I'll close this, delete the branch and come up with something that displays properly via AJAX. I initially noticed it while using get_dialog and sending an incorrect value and noticing the crash.

PR #551, opened 2015-08-05
StringIO to BytesIO changes for portability

frbapolkosnik:
I'd like to change the instances of StringIO to BytesIO for more future proof  implementation.

### DIFF
--- a/crits/core/crits_mongoengine.py
+++ b/crits/core/crits_mongoengine.py
@@ -1,6 +1,6 @@
 import datetime
 import json, yaml
-import StringIO
+import io
 import csv
 
 from bson import json_util, ObjectId
@@ -502,7 +502,7 @@ class CritsDocument(BaseDocument):
 
         if not fields:
             fields = self._data.keys()
-        csv_string = StringIO.StringIO()
+        csv_string = io.BytesIO()
         csv_wr = csv.writer(csv_string)
         if headers:
             csv_wr.writerow([f.encode('utf-8') for f in fields])
--- a/crits/core/data_tools.py
+++ b/crits/core/data_tools.py
@@ -589,7 +589,7 @@ def generate_qrcode(data, size):
     """
     Generate a QR Code Image from a string.
 
-    Will attempt to import qrcode (which also requires Pillow) and StringIO. If
+    Will attempt to import qrcode (which also requires Pillow) and io. If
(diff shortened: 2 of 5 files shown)


mgoffin:
For the sake of being thorough, what is the need to do this? I know we accepted a change for a single service, but I'd like to know the technical advantages of using `io.BytesIO` compared to `StringIO` before we start changing it everywhere. Why not switch it out with `io.StringIO`? What are we gaining/losing here?

frbapolkosnik:
There are several advantages:
- better performance
- portability for Python 3.x
- cStringIO/StringIO.* is going away in 3.x

From what I read io.BytesIO is a drop-in replacement for cStringIO.StringIO, but with StringIO.StringIO one has to be a bit more careful, and that's where io.StringIO comes in. 

As for io.BytesIO over io.StringIO, BytesIO is slightly faster as it operates on bytes rather than unicode according to Python docs: 

https://docs.python.org/2/library/io.html#performance

dicato:
You might want to consider using six with [six.moves](https://pythonhosted.org/six/#module-six.moves) to deal with the `cStringIO` case.

frbapolkosnik:
Thanks, it looks like six just renames cStringIO.StringIO to io.StringIO.

frbapolkosnik:
Looking at:
https://pythonhosted.org/six/#six.StringIO

six.StringIO 
This is an fake file object for textual data. It’s an alias for StringIO.StringIO in Python 2 and io.StringIO in Python 3.
 six.BytesIO 
This is a fake file object for binary data. In Python 2, it’s an alias for StringIO.StringIO, but in Python 3, it’s an alias for io.BytesIO.

frbapolkosnik:
One thing that I've not tested due to lack of access S3 in my setup: crits/core/fields.py 

Tested:
- crits/core/crits_mongoengine.py
I've replaced StringIO.StringIO in crits/core/crits_mongoengine.py with io.BytesIO, and CSV export worked just fine

- crits/core/data_tools.py
generate_qrcode() works properly with io.BytesIO(). I was able to create a valid PNG image that when scaned with my phone, showed the correct data

- crits/emails/handlers.py 
Successfully uploaded Outlook Email (MSG)

- crits/indicators/handlers.py
Successfully uploaded a CSV with indocators

PR #554, opened 2015-08-07
add module to requirements.txt

frbapolkosnik:
totp is using qrcode (crits/core/data_tools.py)


mgoffin:
It's not missing. It's optional.

PR #558, opened 2015-08-12
Updated vocabulary to contain underscores - fix #555

### LINKED ISSUE #555: Issues relationship issues with samples
- On upload of MSG file, the relationship is not forged (or errors out) resulting in attachements and emails not having a relationship
- relationships are also not forged for samples that are containing samples 
- when you try to manually forge a relationship, once you "get the ID from clipboard" the type appears as Email

frbapolkosnik:
This change fixes #555

Looking at the database we have underscores:
{ "_id" : ObjectId("..."), "schema_version" : 1, "forward" : "Contains", "reverse" : "Contained_Within", "active" : "off", "description" : "" }

but https://github.com/crits/crits/blob/master/crits/vocabulary/relationships.py#L16 defines the relationships containing spaces:
CONTAINED_WITHIN = "Contained Within"

### DIFF
--- a/crits/vocabulary/relationships.py
+++ b/crits/vocabulary/relationships.py
@@ -6,54 +6,54 @@ class RelationshipTypes(vocab):
     """
 
 
-    COMPRESSED_FROM = "Compressed From"
-    COMPRESSED_INTO = "Compressed Into"
+    COMPRESSED_FROM = "Compressed_From"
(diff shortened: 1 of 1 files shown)


mgoffin:
The old vocabulary and the new vocabulary are separate and the spaces are intentional.

PR #559, opened 2015-08-12
Made inverse() more tolerant

frbapolkosnik:
Underscores = spaces


mgoffin:
Id rather fix the real issues like your other PR than this bandaid that breaks other things if we really ever want underscores.

frbapolkosnik:
Ok, closing the bandaid PR.

PR #560, opened 2015-08-13
Fix a missed relationship conversion

frbapolkosnik:
"Contained_Within" was missed, this should be the proper fix to get the attachments from emails properly related. There is a PR to fix these things for services https://github.com/crits/crits_services/pull/141


mgoffin:
Is this the one I missed that's causing the email and sample relationship stuff to break?

PR #566, opened 2015-08-17
Updated the field type in email handlers

frbapolkosnik:
Updated the field_type from None to IndicatorTypes.EMAIL_ADDRESS for the "Email To" field.
I think this should make more sense in this case.

### DIFF
--- a/crits/emails/handlers.py
+++ b/crits/emails/handlers.py
@@ -217,7 +217,7 @@ def get_email_detail(email_id, analyst):
                 ))
         email_fields.append(create_email_field_dict(
                 "Email To",
-                None,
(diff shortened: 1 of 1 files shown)


mgoffin:
I don't think this is the right thing to do. The "To" field can be a single email address, or a long list of email addresses, meaning the `IndicatorTypes.EMAIL_ADDRESS` would not always apply. None of them are Indicators either. Saying "any email sent to foo@bar.com is bad" isn't a high quality Indicator. Hence we do not allow making Indicators out of the To field.

frbapolkosnik:
Thanks for the explanation, needed a sanity check there.

PR #568, opened 2015-08-17
First crack at adding filepaths to samples

apolkosnik:
Comments are welcome

### DIFF
--- a/crits/samples/handlers.py
+++ b/crits/samples/handlers.py
@@ -618,6 +618,7 @@ def unzip_file(filename, user=None, password=None, data=None, source=None,
                                              related_type=related_type, backdoor='',
                                              user=user, campaign=campaign,
                                              confidence=confidence,
+                                             filepath=filepath,
                                              bucket_list=bucket_list,
                                              ticket=ticket,
                                              inherited_source=inherited_source,
@@ -646,7 +647,7 @@ def unrar_file(filename, user=None, password=None, data=None, source=None,
                method="Generic", reference='', campaign=None, confidence='low',
                related_md5=None, related_id=None, related_type='Sample',
                bucket_list=None, ticket=None, inherited_source=None,
-               backdoor_name=None, backdoor_version=None):
+               backdoor_name=None, backdoor_version=None, filepath=None):
     """
     Unrar a file.
 
@@ -744,6 +745,7 @@ def unrar_file(filename, user=None, password=None, data=None, source=None,
                                                      campaign=campaign,
                                                      confidence=confidence,
                                                      bucket_list=bucket_list,
+                                                     filepath=filepath,
                                                      ticket=ticket,
                                                      inherited_source=inherited_source,
                                                      relationship=relationship,
@@ -767,7 +769,7 @@ def handle_file(filename, data, source, method='Generic', reference='', related_
                 related_id=None, related_type='Sample', backdoor=None, user='',
                 campaign=None, confidence='low', md5_digest=None, bucket_list=None,
                 ticket=None, relationship=None, inherited_source=None, is_validate_only=False,
-                is_return_only_md5=True, cache={}, backdoor_name=None,
+                is_return_only_md5=True, cache={}, backdoor_name=None, filepath=None,
                 backdoor_version=None):
     """
     Handle adding a file.
@@ -817,6 +819,8 @@ def handle_file(filename, data, source, method='Generic', reference='', related_
     :type backdoor_name: str
     :param backdoor_version: Version of the backdoor to relate the file to.
     :type backdoor_version: str
+    :param filepath: The filepath of the file.
+    :type filepath: str
     :returns: str,
               dict with keys:
               "success" (boolean),
@@ -873,10 +877,13 @@ def handle_file(filename, data, source, method='Generic', reference='', related_
         is_sample_new = True
         sample = Sample()
         sample.filename = filename or md5_digest
+        sample.filepaths.append(filepath)
         sample.md5 = md5_digest
     else:
         if filename not in sample.filenames and filename != sample.filename:
             sample.filenames.append(filename)
+        if filepath not in sample.filepaths:
+            sample.filepaths.append(filepath)
 
         if cached_results != None:
             cached_results[md5_digest] = sample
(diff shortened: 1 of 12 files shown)


crits/samples/handlers.py
```diff
@@ -1443,6 +1459,31 @@ def update_sample_filename(id_, filename, analyst):
     except ValidationError, e:
         return {'success': False, 'message': e}
 
+def update_sample_filepaths(id_, filename, analyst):
+    """
+    Update a Sample filename.
+
+    :param id_: ObjectId of the Sample.
+    :type id_: str
+    :param filename: The new filename.
```
mgoffin: Just a copypasta error.

crits/samples/handlers.py
```diff
@@ -1443,6 +1459,31 @@ def update_sample_filename(id_, filename, analyst):
...
+    :type filename: str
+    :param analyst: The user setting the new filename.
```
mgoffin: Just a copypasta error.

crits/samples/templates/samples_detail.html
```diff
@@ -75,6 +76,16 @@ <h3 class='titleheader'>
                 </td>
             </tr>
             <tr>
+                <td class="key">Filepaths</td>
```
mgoffin: I noticed you have a `filepath` attribute as well as a `filepaths` attribute (I assume for the same reason we did this for filename..so there's one main filepath you want to refer to, with references to other places it could be). I don't, however, see the `filepath` exposed in the details template?

documentation/help_main.txt
```diff
@@ -1047,6 +1047,7 @@ information, as follows:
 
 :Upload type: Choose whether the sample is a file or metadata
 :Filename: The name of the file encountered
+:FIlepath: The path of the file encountered
```
mgoffin: Looks like the `I` needs to be lowercase. I don't see this error reflected in the `help_main.html` file which makes me think you did not use `rst2html.py` to re-generate the `help_main.html` file. You should never edit the html file by hand. Always re-generate it off the txt file.

mgoffin:
The only other thing I'm mulling over outside of the above is the addition of the Indicator and Object vocab. The `File Name` indicator was something I considered "double-duty" for both file names and full file paths. Chatting with some folks to see if we can come to a consensus about splitting it out like you did.

crits/samples/templates/samples_detail.html
frbapolkosnik: My intention was to have a single filepath entry associated with a submitted sample, but populating the filepaths list.

crits/samples/handlers.py
frbapolkosnik: I think this function might be not needed here, I'll have to check.

mgoffin:
Makes sense, but like file name it would be nice to get to choose and visually see which one you've selected for your file path and to be able to change it whenever you want instead of it being clumped into the list.

mgoffin:
Also, about the File Path indicator type. I've talked to a few people about having three distinct types:

- File Name: just the name of a file
- File Path: full path and file name
- Directory: just the full path to the directory where a file lives

Would this make sense? I've gotten positive feedback so far. It would mean we can take the File Paths for a sample and have a "create indicator" feature which can either create a Directory indicator out of the File Path, or a File Path Indicator by combining a File Path and a File Name. But it also begs the question of renaming "File Path" for Samples into "Directories" to make that a more logical transition?

frbapolkosnik:
I've stepped across this
https://www.blackhat.com/docs/us-14/materials/us-14-Castle-GRR-Find-All-The-Badness-Collect-All-The-Things-WP.pdf

Then I started thinking about windows path separators vs the rest of the world.

In order to have something useful there I think that having the full path where the file was found, would be great. Then someone will ask for globbing, and then for regex.

For regular web based interface access patterns passing the absolute paths through os.path.basename(), os.path.dirname(), os.path.splitext() should be fine.

My perception is that the pre-chopped forms might be useful for pivoting and querying, but then we'd end up duplicating some data. Perhaps the lowest overhead solution would be to pre-split, then use os.path.join(). Directory and filename make sense, and the only ambiguity will happen when you'll end up with many directories and many filenames, and you'd want to create some actionable indicators, but let's say the filename is dependent on the directory name, and not having connection between filename and directory could be problematic in a sense that it could increase the number of combinations e.g.:
when lots of filenames are occurring in in directory foo, but only few in bar and baz, the pre-chopped model will generate lots of variants that might not exist.

For now I mostly rely on CRITs to hold the data for the observed artifact feeds (sandboxes, AV), and there's not much variety when it comes to filepaths.

crits/samples/urls.py
```diff
@@ -20,5 +20,7 @@
         (r'^yarahits/list/$', 'yarahits_listing'),
         (r'^yarahits/list/(?P<option>\S+)/$', 'yarahits_listing'),
         (r'^set_filename/$', 'set_sample_filename'),
+        (r'^set_filepath/$', 'set_sample_filepaths'),
         (r'^filenames/$', 'set_sample_filenames'),
```
frbapolkosnik: I think that one of the two can be removed here.

crits/samples/sample.py
```diff
@@ -63,7 +64,9 @@ class Sample(CritsBaseAttributes, CritsSourceDocument, Document):
 
     filedata = getFileField(collection_name=settings.COL_SAMPLES)
     filename = StringField(required=True)
+    filepath = StringField()
```
mgoffin: What is this for if it's not something that can be set?

frbapolkosnik:
Got some more fixes done, even email attachments should work now

frbapolkosnik:
Do you want it to look like the edit fields for the email properties?

mgoffin:
When submitting a zip file (and I assume a rar file) the filepath in the form isn't properly passed along so it is dropped on the floor.

frbapolkosnik:
if you get "2 No such file or directory" then the path for 7z is not right. Brew puts it in /usr/local/bin/7z, whereas your default setting might be /usr/bin/7z.

mgoffin:
A head's up about the status of this PR: Still waiting for a response to my above comment about filepaths being dropped on the floor for archive files. Not sure about the best solution:

- filepath for exploded archive file is added for each exploded file?
- filepath plus relative path inside of archive file is used for each file in archive?

frbapolkosnik:
I think that I'll just add a runtime switch

frbapolkosnik:
I'll need to add the logic for this to work.

frbapolkosnik:
This one should be ready for testing, and perhaps some naming changes.
One tiny caveat is that I left the RAR path untouched, as #646 removes that part.
Logic added for "inherit_filepath" and filepath should be passed along just fine now. Also, I've changed the parameter for archive extraction, so now we can get the relative paths. I welcome suggestions for a better caption/name for "inherit filepath", it doesn't feel really intuitive. Perhaps "add paths from archive"

PR #579, opened 2015-08-26
Fix infinite recursion bug in type filter and return a string as documented

lelonek1:
This fixes the type filter calling itself rather than the built-in type function.
The type is also converted to a string before being returned to match the documented return type.

### DIFF
--- a/crits/core/templatetags/filters.py
+++ b/crits/core/templatetags/filters.py
@@ -74,8 +74,8 @@ def does_field_have_indicator(field, relationships):
 
     return return_value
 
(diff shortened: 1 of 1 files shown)


mgoffin:
Would this need to get fixed in any templates that are using it? Otherwise this will cause template errors (I don't know if it is actually being used).

lelonek1:
It is not being used by any templates that I could find (other than my own), but if there were any they would not be working as expected.  I specified the name in the decorator so that if there were any templates using it, they would not need to be modified.

PR #580, opened 2015-08-26
Fix creating initial admin account in bootstrap script

lelonek1:
The initial admin account created by the bootstrap script must be set as active for the user to log in.
Otherwise, the user would need to run `python manage.py users -s <username>` before they could log in.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -167,7 +167,7 @@ then
   read -p "Organization name: " ORG
   
   echo "Adding a default admin account"
-  python manage.py users -a -A -e "$EMAIL" -f "$FIRSTNAME" -l "$LASTNAME" -o "$ORG" -u "$USERNAME"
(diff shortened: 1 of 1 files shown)


mgoffin:
The default value for is_active is True and it is not set to False anywhere when creating a new user. How were you triggering this? Also if this is repeatable for every new user account the fix should go into the users management command and not into the bootstrap.

lelonek1:
I experienced this each time I ran the bootstrap script on a clean install of Ubuntu 14.04 Desktop.  

I just tested adding a user to my current installation and it was created as active.  I am not sure why I was experiencing different behavior when running the bootstrap script.
I also just tested running the bootstrap script on clean clean copy of the code from master and it created the first user as active as well. 

It was quite annoying when I was first trying to learn about and setup CRITs because I would run the bootstrap script but could not log in as the user it created.  It took a lot of trial and error to find that the initial user was not being set as active, but now I cannot seem to replicate that behavior.  I do not currently have access to the system on which I was originally testing CRITs, so I will have to let you know later if I can replicate the issue again there.

mgoffin:
I'll close this for now. If you can replicate again in the future let us know!

PR #585, opened 2015-09-02
First cut at Service provisioning in Vagrant

gtback:
Make it easier to have some services installed by default in a Vagrant-provisioned dev environment. The next step would be to automatically install dependencies for services that need them, so those services work too.

cc @packet-rat

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,6 +6,14 @@ Vagrant.configure(2) do |config|
 
   config.vm.network "forwarded_port", guest: 8080, host: 8080
 
+  # Assumes that, if present, the "crits_services" repository is checked out
+  # beside the "crits" repository.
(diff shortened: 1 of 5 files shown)


mgoffin:
Neat. So any services whose dependencies are satisfied with the base dependencies for CRITs should be setup, and anything else just doesn't show at all?

gtback:
Set up pending any necessary configuration options, yes.

frbapolkosnik:
And it broke half of my services ;-)

crits.service.core Failed to import service (some_service): cannot import name settings

where some_service: virus_total, chminfo, upx, peinfo, cuckoo, office_meta, carver, unswf

mgoffin:
Submitted some fixes to master to take care of the issues.

PR #593, opened 2015-09-10
Bootstrap fixes for CentOS/RHEL

frbapolkosnik:
I took some niceties from #510 and #511, and added some other fixes. Tested on a fresh install of CentOS 6.


mgoffin:
I think this is a good route to take. I'll accept and we can iron out bugs as we all find them.

gtback:
This breaks installation on Ubuntu, since the `python-pip` package does not install an executable named `pip2.7`

PR #604, opened 2015-09-18
Update bootstrap to fix pip issue on Ubuntu

frbapolkosnik:


gtback:
:+1: This seems to fix it in Vagrant/Ubuntu

Issue #586, opened 2015-09-03
Clicking the + next to fields in an email adds strings instead of the field type


tekneke:
When attempting to press the +(Create Indicator) next to a field in an email the indicator type is a "string" as opposed to the field type itself i.e. X-mailer, Subject, etc. Wasn't if this was as intended. Also, is there a reason that if an indicator already exists that the email can't be related to it?

PR #607, opened 2015-09-22
Email Create Indicator Fix Plus Sign (+)

thelok:
When on a email details page, clicking the plus sign (+) normally creates an indicator. This was broken because in emails/views.py the indicator types were being set to the following types which are not in the vocabulary anymore. So when you try to create an indicator it complains that it is not a valid indicator type.

```
               if header_field in ("from_address, sender, reply_to"):
                   ind_type = "Address - e-mail"
               elif header_field in ("originating_ip", "x_originating_ip"):
                   ind_type = "Address - ipv4-addr"
               else:
                   ind_type = "String"
```

I updated the code to use the new vocabulary types by putting the types in the correct locations in the email_details.html template and updating the back end to also grab the new field.

I strongly suggest reviewing the code to make sure nothing is wrong because I haven't done much with the indicator stuff after the removal of STIX/CyBOX.

### DIFF
--- a/crits/core/crits_mongoengine.py
+++ b/crits/core/crits_mongoengine.py
@@ -1587,6 +1587,7 @@ class CritsBaseAttributes(CritsDocument, CritsBaseDocument,
         my_rel.object_id = rel_item.id
         my_rel.rel_confidence = rel_confidence
         my_rel.rel_reason = rel_reason
+        is_left_rel_exist = False
 
         # setup the relationship for them
         their_rel = EmbeddedRelationship()
@@ -1598,6 +1599,7 @@ class CritsBaseAttributes(CritsDocument, CritsBaseDocument,
         their_rel.object_id = self.id
         their_rel.rel_confidence = rel_confidence
         their_rel.rel_reason = rel_reason
+        is_right_rel_exist = False
 
         # check for existing relationship before
         # blindly adding them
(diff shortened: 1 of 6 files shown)


thelok:
Fixes #586

thelok:
OK so I made another change, while taking a look at issue #586 I ran into the issue where I couldn't create a relationship because I got into a situation where only the "right side" had a relationship (i.e. unidirectional relationship).

Just because you have a unidirectional relationship doesn't mean you should fail to create a relationship, just create the other side of the relationship. 

Anyways, take a look at this, if you decide not to take this specific fix then I have no problem with backing out the change but I will be keeping this fix in my repo.

thelok:
One last note about the unidirectional relationships... I was able to recreate my issue by very rapidly clicking all 10 plus signs (+) on the email details page. There is some sort of race condition where some of the relationships on the left side are overwritten or are not created, therefore you get the situation where you have a unidirectional relationship instead of the bidirectional relationship. I'm not going to fix this issue here.

brlogan:
I think it might be better to repair the broken relationship by grabbing the data from the existing half, rather than adding a new half that doesn't match the other half. At least, that's the way I handled it. Thoughts?

frbapolkosnik:
Working on a PR already

frbapolkosnik:
The indicator_from_header_field() was assigning old IndicatorTypes, I've converted it to the new style Vocabulary, works like a charm.

thelok:
@brlogan I'm not opposed to copying the data from the other half of the relationship as long as you can assume that half is 'correct'

@frbapolkosnik Actually my fix already takes care of this, the client side code passes in the indicator type. The types are already defined in emails/handlers/get_email_detail() in the "email_fields" variable.

brlogan:
The existing half should be the most correct. It will have the analyst that originally formed (attempted to form?) the relationship, and the original date. It may be important to maintain this history, rather than muddying it by having different halves.

thelok:
Ok, pushed up some new logic to copy the existing relationship's data if only half of the relationship is found.

seanthegeek:
Just tested it. All headers worked except reply-to.

frbapolkosnik:
Well, trying to add an email in the form of "Test2 <test2@example.com>" will not work...

mgoffin:
Because it's not a valid email address.

frbapolkosnik:
\"Test2 <test2@example.com>\"  ?

frbapolkosnik:
\"Test2 \<test2@example.com\>\"

seanthegeek:
See PR #612

mgoffin:
Still not a valid address?...

PR #608, opened 2015-09-22
This should fix adding the indicators from email

frbapolkosnik:
The issue was mentioned in #607... reason, ind_type passed to handle_indicator_ind() was not using the new vocabulary, and it was bombing out in validate_indicator_value()


frbapolkosnik:
@brlogan I think this is what you meant in your comment for #607. Anyway, I think that this is the proper fix.

brlogan:
I think this is something different. I was talking about the handling of broken, half-formed, relationships.

PR #614, opened 2015-10-06
Speed up adding results

trogdorsey:
In _add_result, instead of doing a loop to update the dict,just use the builtin .update() to do the same thing

Also added _add_results function to speed up adding large amounts of data.  I have some test samples that add a lot of data, and it was taking > 3 hours for it finish.  Using _add_results, it took 7 seconds.


mgoffin:
I'm down for speed increases of 1543%!

PR #620, opened 2015-10-14
Added possibility to add results in batch via the api

cynexit:
**Issue**
The service api already exposes the possibility to add results to a TLO which is very important for distributed services. However, a distinct HTTP request is necessary for each result set which puts a lot of unnecessary extra strain on both CRITs and the distributed service.

**Proposal**
By adding the option to send result sets in batch as JSON lists it would be possible to handle many result sets on both sides with a single HTTP request, eliminating a lot of overhead.

**Improvements**
During our internal testing we pushed 5000 result sets to 100 TLOs each on a CRITs machine on the same network. Depending on the size of the result sets we have seen a 20% to 50% speed improvement. We assume the improvement would be even greater for machines not on the same network.

**Known Issues**
None so far, our update doesn't break backwards compatibility and doesn't add any extra strain to non-batch requests.

### DIFF
--- a/crits/services/api.py
+++ b/crits/services/api.py
@@ -1,3 +1,4 @@
+import json
 from django.core.urlresolvers import reverse
 from tastypie import authorization
 from tastypie.authentication import MultiAuthentication
--- a/crits/services/api.py
+++ b/crits/services/api.py
@@ -55,6 +55,7 @@ class ServiceResource(CRITsAPIResource):
         result = bundle.data.get('result', None)
         result_type = bundle.data.get('result_type', None)
         result_subtype = bundle.data.get('result_subtype', None)
+        result_is_batch = bundle.data.get('result_is_batch', False)
         log_message = bundle.data.get('log_message', None)
         log_level = bundle.data.get('log_level', 'info')
         status = bundle.data.get('status', None)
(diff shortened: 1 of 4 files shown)


mgoffin:
This seems promising, however a bit confusing from a usage perspective. It's not very clear unless you read the code that you are supposed to make each parameter an equal-length, properly ordered list to match with each other.

One thing that would help is to update the API documentation found here:

https://github.com/crits/crits/blob/master/documentation/help_main.txt

Once that is updated in the appropriate spot (somewhere in here: https://github.com/crits/crits/blob/master/documentation/help_main.txt#L1834), you'll need to run the following:

```zsh
$ cd documentation
$ python rst2html.py help_main.txt help_main.html
```

That will properly update the html file which is used in the UI for the Help page without you having to manually edit it. It will also tell you if there's any issues with the changes you made so you can fix them.

cynexit:
Thanks for pointing me to this, I updated the files. (I thought the complete documentation is handled trough the github wiki pages so I planed to update these, but this is actually a much nicer way)

cynexit:
Another major performance gain: In the previous versions we looped over the results and added each result via add_result(). This is the way the api would do it in a normal scenario, but this means many update commands for mongodb. Since each update locks mongo up this procedure was painfully slow when adding large datasets.
As a conclusion we added add_results() which takes lists instead of strings, prepares one big list and updates the results using push_all with one update request.
This cut our time of submitting 5000 analysis results from ~10m to ~1s so I'm shearing it here. We had to modify handlers.py as well as api.py but everything is backwards compatible.
This way adding results via the api is more or less as fast as adding them trough a local service.

PR #624, opened 2015-10-15
install M2Crypto for ubuntu 14.04 server

cjarruda:


frbapolkosnik:
That is already covered by #622

PR #627, opened 2015-10-22
New actor identifier form broken

thelok:
Try to create a new actor identifier:

    items=>actors=>actor identifiers=>new actor identifier

Changed it so that actors form for creating new actor identifiers use the new vocabulary instead of trying to look for it in the DB where the table doesn't exist anymore.

### DIFF
--- a/crits/actors/forms.py
+++ b/crits/actors/forms.py
@@ -1,13 +1,15 @@
 from django import forms
 from django.forms.util import ErrorList
 
-from crits.actors.actor import ActorThreatIdentifier
 from crits.campaigns.campaign import Campaign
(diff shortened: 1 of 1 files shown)


mgoffin:
Actor Identifiers are not part of the vocabulary. Actor Identifier Types are also not part of the vocabulary.

You add new Actor Identifier Types so when you go to add a new Actor Identifier you can choose from the dropdown what Identifier Type it is.

Both of these are driven by the database so this change should not be made.

thelok:
Ugh, nevermind. You can create Actor Identifier Types under Admin=>Add New Item. So this is different than the vocab types.

PR #628, opened 2015-10-22
Replacing m2crypto with pycrypto

frbapolkosnik:
I don't use totp, so I need somebody to test it.

### DIFF
--- a/crits/core/totp.py
+++ b/crits/core/totp.py
@@ -1,6 +1,7 @@
 import hmac, base64, struct, hashlib, time
 from django.utils.crypto import pbkdf2
-from M2Crypto import EVP, Rand
+from Crypto import Random
+from Crypto.Cipher import AES
 
 # from http://stackoverflow.com/questions/8529265/google-authenticator-implementation-in-python
 def get_hotp_token(secret, intervals_no):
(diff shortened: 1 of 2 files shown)


frbapolkosnik:
Just a quick example taken from 
http://crypto.stackexchange.com/questions/24050/why-dont-pycrypto-and-m2crypto-produce-the-same-result-as-openssl-here-aes-128

```
from Crypto.Cipher import AES
key = b'YELLOW SUBMARINE'
cipher = AES.new(key, AES.MODE_ECB, "")
msg =  cipher.encrypt(b'0123456789abcdef0123456789abcdef')
repr(msg)
```

```
import M2Crypto
cipher=M2Crypto.EVP.Cipher('aes_128_ecb',b'YELLOW SUBMARINE',"", op=1,
padding=0, salt=None)
out=b''
plaintext=b'0123456789abcdef0123456789abcdef'
out=out+cipher.update(plaintext)
repr(out)
```

frbapolkosnik:
I can try to re-implement this with Cryptography (https://cryptography.io/en/latest/) if that's a better option for redoing the taxii_service.

On the services side, the taxii_service is the only one that used M2Crypto.

frbapolkosnik:
The reason for this is that M2Crypto doesn't seem to be moving anywhere (Support for Python 3.x is not there, and we're having issues building it through pip).

wzod:
FWIW I ran into a similar issue with building M2Crypto using pip just recently (had not observed that issue before).  I was able to successfully resolve the issue using python-m2crypto, and so far no significant issues.

Above suggestion might be a short-term fix as switching to pycrypto makes more sense.  I'm currently not using totp, but I will certainly test as soon as I have some cycles.

frbapolkosnik:
In CRITs core this is the only spot (TOTP) where m2crypto is used. I think it was said that TOTP is optional, but still the m2crypto is a requirement... Since the taxii was split out as a service, there is no need to keep m2crypto as a requirement. In CRITs services the only service that uses m2crypto is taxii_service. I suppose that since m2crypto has issues building, this is a sensible workaround to get the core usable by removing this dependency. I'll add an entry for UPDATING if I'll more people think this is a good idea.

frbapolkosnik:
After some more looking around I've decided that I'll close this PR and create a new one that uses cryptography module instead. With that It'll be much easier to migrate TAXII_service too.

PR #629, opened 2015-10-25
libjpeg-dev and python-m2crypto are required on Ubuntu

jedisct1:
The former is required for PIL.

The later is required because it doesn't compile out of the box on Ubuntu.

These additions make the Vagrant image actually work.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -248,7 +248,7 @@ ubuntu_install()
     echo 'deb http://downloads-distro.mongodb.org/repo/ubuntu-upstart dist 10gen' | sudo tee /etc/apt/sources.list.d/mongodb.list
     sudo add-apt-repository universe
     sudo apt-get update
-    sudo apt-get install -y --fix-missing build-essential curl git libevent-dev libz-dev libfuzzy-dev libldap2-dev libpcap-dev libpcre3-dev libsasl2-dev libxml2-dev libxslt1-dev libyaml-dev mongodb-org numactl p7zip-full python-dev python-pip ssdeep upx zip swig libssl-dev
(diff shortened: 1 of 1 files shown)


mgoffin:
Technically Pillow is the requirement for the image library, and this is only one of like a dozen optional dependencies for image format support.

jedisct1:
We really should just add `python-pillow` to the dependencies, then. It's automatically pulling `libjpeg-turbo` and other libraries requires for image format support.

mgoffin:
https://github.com/crits/crits/blob/master/requirements.txt#L2 

Does that not cover it?

mgoffin:
We try to stay away from installing python dependencies via anything other than `pip` since developers of libraries tend to be more up-to-date than waiting for OS package managers to keep up with them.

frbapolkosnik:
I've ran into this too, but the issue is that perhaps it's cleaner from the multi-platfrom perspective to install python libraries through pip, and anything else through the native package managers. Also this approach will be rather clean if/when the move to Python 3.x happens.

mgoffin:
We already install python libraries through pip and everything else through native package managers...

jedisct1:
`m2crypto` doesn't properly compile using `pip` on Ubuntu.

And for `Pillow` to compile using `pip`, `libjpeg-dev` is the bare minimum to have.

apolkosnik:
#683 should fix the Pillow issue for Debian, Ubuntu, RedHat, and CentOS. As for the m2Crypto, there is an error message that describes how to get it fixed, also there's a patch to replace it in the core with PyCrypto #628, though the taxii_service is still using m2crypto, but then taxii_service is not working at this time. Choices, choices...

PR #632, opened 2015-10-28
Fix test suite (Dashboards, IPs, Relationships)

dreardon:
The test suite hasn't been working for a while due to vocabulary and default dashboard creation changes. I am creating some tests for the new Signature TLO and was hoping to get this in there so that we had a working baseline.

I believe that the only "controversial" aspect of this commit might be the way that I'm muting the standard output when creating the default dashboard. Basically, I wanted to use crits.core.management.commands.create_default_dashboard to still do the creation, but the standard out in that call would possibly cloud up the test results.


mgoffin:
Couple things I noticed:

- If we are using the new vocabulary, we should be using the vocabulary classes, not static strings
- I am not a fan of dropping collections that can actually have useful data in them. It means once you populate the system you can never run the test suite anymore. If you are adding test content to real collections, we should keep track of what's added and specifically remove it.

mgoffin:
I know it's nit-picking, but wouldn't it have been easier to set `IP_TYPE` and `TRELATIONSHIP_TYPE` to the new vocabulary instead of creating a new variable that's embedded in the function?

dreardon:
Don't worry about nit-picking, I'm definitely up for learning different ways to do things. You're right and it probably would have been easier to do that; however, my reasoning for embedding it in the function was in case we wanted to use different IP_TYPE vocabulary in different functions. Yes, I could have then defined two variables at the top for IP_TYPE_V4 and one for IP_TYPE_V6, but I didn't go that route. It's also my reasoning for TRELATIONSHIP_TYPE.

apolkosnik:
Some of the changes proposed here are already merged, you might want to update the pull. Thanks!

PR #638, opened 2015-10-30
Add preferred list to Actions.

wxsBSD:
When creating a new IndicatorAction (this will be a misnomer in the
future, more on that later) you can now select which TLOs this Action is
"preferred" for. An action having "preferred" status means that from the
details page or the listing page you can click a single button and all
actions that are preferred for that TLO type will be applied.

Currently the only one that this matters for is Indicators, but the code
has been built so that when Actions become an attribute of all TLOs the
same interface can be used.


crits/core/handlers.py
```diff
@@ -1230,6 +1230,69 @@ def toggle_item_state(type_, oid, analyst):
...
+
+    actions = []
+    now = datetime.datetime.now()
+    # Get preferred actions and add them.
+    for a in preferred_actions:
+        action = {'action_type': a.name,
+                  'active': 'on',
+                  'analyst': username,
+                  'begin_date': now,
+                  'end_date': now,
```
mgoffin: I might just leave the end date as None (or whatever it is that leaves it open-ended)? It would seem odd that an Action would begin and end at the exact same time.

wxsBSD: Not sure why I didn't try None when I was dealing with that earlier. I was using ```''``` and it would stick in the default value (```datetime.now()```) instead. I've switched it to ```None``` now.

thelok:
How far along are you with moving IndicatorAction out of indicator? I might provide you some code if you haven't started it.

wxsBSD:
I haven't started it, nor do I have any free time coming up to work on it. If there isn't already an open issue for it I would make one and start tracking design and goals there.

mgoffin:
I am starting to work on it. It seems the feedback from the preferred actions is that it's not usable in its current state. So I am going to redo it along with the move to make Actions separate.

If you have some code I can look at I'd take it :)

cdorer:
Nice, so it sounds like Actions will be a separate collection?

thelok:
Why is preferred actions not usable? Is it because there's no customization for default values?

mgoffin:
Because they are based only on TLO type, meaning the same Actions you would prefer for an IP Indicator are the same ones you'd prefer for an Email Address, which is really not the case at all.

The Preferred Actions need to be able to specify a list of dictionaries it prefers. Each dictionary containing and Object Type (Indicator, Sample, etc.), Field to look at, and Value to look for in that field.

thelok:
I pushed up a branch: 

https://github.com/thelok/crits-1/tree/initial_generalize_actions

You can review and merge these changes into your branch and continue working from there. I don't intend to finish the rest of the work... I'll leave it up to you guys if you want to change anything.

Diffs here: https://github.com/crits/crits/compare/master...thelok:initial_generalize_actions?expand=1

The biggest changes is that actions operate on obj_type and obj_id so it can apply to any TLO. Also I widget-ized the actions row widget thingy so that it can be included into any TLO details page.

Lastly there is a new base class CritsActionsDocument in crits/core/crits_mongoengine.py that a TLO needs to extend to get generalized Action functionality. I made Indicator extend this base class so that it Actions still works like before.

PR #646, opened 2015-11-09
Remove unrar

frbapolkosnik:
I'm removing unrar since there is no longer need to use it because 7z already includes unrar functionality.


frbapolkosnik:
This one is ready for testing.

Please note that besides removal of the Unrar button also the sample upload form has been updated to remove RAR choice. From my limited testing, both the sample uploads and Unzip (renamed to Un7zip) are working properly with both ZIP and RAR archives.

Issue #625, opened 2015-10-15
Signature Identification/Storage


dreardon:
I wanted to follow-up on a part of an old post from June 2015 and request a way to handle/store signatures:
https://github.com/crits/crits/issues/503#issuecomment-108079377

That discussion was more based on the CRITs vocabulary, but I’m looking for the best place or method to store signatures that have been developed. I guess I see a few different ways of going about it, but I wanted to approach the community to hear how others might be handling it, or if it should be considered as new development. Options I see are:

 - Start the process to make Signatures a new TLO, realizing that this is not trivial work
 - Use the “Raw Data” TLO. If I put it there, I can also identify the tool (Snort, Yara, etc.)
 - Change the vocabulary in the Indicators TLO to define/categorize an indicator...something that the June 2015 suggested against

Thanks.

mgoffin:
I still think the third option is out. Signatures to me are not Indicators of Compromise, they are used as detectors of lots of those Indicators as well as other malicious content.

The second option isn't the worst thing in the world as it would cater to the immediate need, but I still think the first option is the best. It gives us a specific place for all signatures/rules to live and makes it easier for building services around them.

dreardon:
Thanks for the response. I’ve done *some* development, but before I go too far, I wanted to pass along my thoughts on the Signature TLO to see if anyone has any thoughts or suggestions for changes. My approach is to model it after the Raw Data TLO with some differences. Essentially, I kept it pretty simple on purpose. The data elements I wanted to capture in the Signature form are:

- Title
- Data Type (Administrator managed drop-down of types…Yara, Snort, etc.)
- Description
- Data
- *Source*
- *Source Method*
- *Source Reference*
- *Bucket List*
- *Ticket*

The items in italics above are the standard form elements found in most TLOs. The Signature TLO will also contain a similar versioning and diffing capability that exists in the Raw Data TLO.

![image](https://cloud.githubusercontent.com/assets/346206/10637440/bdb96122-77d3-11e5-9085-5bb3eace3139.png)

Finally, the Signature API will be pretty standard and adhere to what the rest of the TLOs allow:

![image](https://cloud.githubusercontent.com/assets/346206/10637422/9983cae0-77d3-11e5-88e0-179c7ad8f7ab.png)

mgoffin:
This looks great! I think you've hit on all of the appropriate fields
necessary to build up a solid TLO. Has there been any ideas on how to
display the "data" field on the details page? For example, Yara rules are
[...]

PR #655, opened 2015-11-17
New Signature TLO

dreardon:
This PR should address all the changes that were discussed in CRITs request https://github.com/crits/crits/issues/625. The features include:
 - A new TLO for Signatures which was modeled after the Raw Data TLO
 - New meta-data for Tool Type, Min/Max Versions of the tool, and Tool Dependencies
 - Tool Dependencies are a type-ahead list of values, but still allows new entries (similar to Buckets)
 - Instance-specific list of available Signature Tools defined by the admin (Bro, Snort, and Yara added by default)

After this is worked out and accepted, there are plans to add new services which will take advantage of this new TLO. Updates could also be made to existing services to take advantage of this TLO, but those could be discussed after we have this as part of the baseline.

Thank you very much.


dreardon:
Good point about better emphasizing the signature itself. I didn't want the "weight" of the line by line commenting and highlighting of the raw data "data" component for this purpose, but I think I might have gone too far in the other direction of just making it another data element on the page and it blends in. I'll see what I can do.

mgoffin:
Yea I agree that removing the raw data view was a good idea. It would be overkill for sure. Thanks for looking into this :) After that, if everything works in the UI and the code looks solid, I suspect we'll get this accepted asap!

dreardon:
Thank you both for the quick feedback. I made some adjustments based on your input and added them to this PR (see above). Please let me know if you notice anything else. Thank you again.

mgoffin:
I suspect you are using the diffing piece wrong. You need to select TWO things to diff. You are most likely only selecting one thing and assuming it's diffing against the one you are viewing, which is incorrect.

mgoffin:
I think this looks good! I'm going to accept it into master and make one quick change after. Since there's no real difference between the `Data` section where the signature is displayed and the contents of the `Signature` tab, we might as well remove the `Signature` tab and make some room for potential services. Once it's in master we can get a lot of eyes on it and get some more testing in to polish stuff up.

Thanks for doing this! I can see this being amazingly useful and once services are written for them it will really change things for a lot of people!

armtash:
Have you considered adding an "Activity" section to the Signatures TLO similar to how one exists for Indicators?  It would be a nice addition so that one can track hits of different signatures.

PR #657, opened 2015-11-19
Actions for all TLOs

mgoffin:
This adds Actions as a feature to all TLOs.

Gives Actions the ability to specify which TLO(s) the Action applies to.

Allows you to specify a list of conditions making that Action "preferred". This is a list of CSVs like so:

```
Indicator, ind_type, Domain
Email, to, foo@bar.com
```

The above example would limit this Action to being preferred only to Indicators with a type of Domain and Emails where foo@bar.com is in the To list.

NOTE: Any existing Actions need to be updated with the list of TLOs they apply to otherwise they won't show up in the dropdown list when adding an Action.


thelok:
I see you kept actions as an embedded object instead of its own TLO like the Analysis Results, any reason why?

mgoffin:
It turned out to be too much work all at once. Seemed to make sense to at least get this going before trying to make another drastic change/migration.

PR #664, opened 2015-12-01
API Enhancement

armtash:
This might be longer than necessary so I apologize in advance.

This PR is in reference to a google groups discussion located [here]( https://groups.google.com/forum/#!topic/crits-users/P4CZQCGTTLA) .

The purpose of these changes is to expand the API, using "PATCH" requests which were enabled in https://github.com/crits/crits/pull/449

The following core functions are exposed with these changes:

Core:
status_update
ticket_add
ticket_update
ticket_remove
source_add_update
source_remove
action_add
action_update
action_remove
description_update

Indicator:
set_indicator_attack_type
set_indicator_threat_type
activity_add
activity_update
activity_remove
ci_update (Update confidence or impact)

I followed the following guidelines:

1.  Only "views.py" and "handlers.py" files are modified.  No "core" functions are modified, such as "crits.core.crits_mongoengine.py" or "crits.indicators.indicator.py" (We may need to rethink this, see notes below)

2.  "crits.core.api.py" is modified to expose handler functions.

3.  Each handler function should be passed "id_", "type_", "user", and "***kwargs" at a minimum.  Functions that use similar arg names are renamed ("object_id" -> "id_", "analyst" -> "user") ***kwargs is necessary to account for additional arguments which are passed to these functions.  
4.  Any function with "action" as an argument needs to be renamed.  This conflicts with the "action" that is passed in an API call.
5.  Since all of these handlers require datetime.datetime objects instead of strings, these need to be validated before performing any actions at the handler level, since only strings are being passed from API calls.  I modified datetime_parser to check and modify date strings or dicts which contain date strings: https://github.com/armtash/crits/blob/api_patch/crits/core/handlers.py#L1354.  

Here are the potential issues:

1.  As I mentioned, I avoided modifying the core functions, however I think it may be necessary to do so for some of these functions that are exposed.  The main reason would be to raise/return an error when an action is not successfully completed.  For example, https://github.com/crits/crits/blob/master/crits/core/crits_mongoengine.py#L259 As you can see, in the set_status function, a change occurs only if a valid status is passed.  However, no error is raised if an invalid status is passed.  You can work around this by performing checks in the handlers themselves.  If the intended change was not successfully saved, raise an error.  

2.  I have not done any error checking/validation in the handlers aside from https://github.com/armtash/crits/blob/api_patch/crits/core/handlers.py#L372.  This means that if the user attempts to make a change which is not successful, they are still returned a "success" message.  As you can imagine, this is very misleading.

3.  For some functions, a "date" is passed which is used as the key for an instance of a particular attribute.  For example, "crits.core.source_add_update" https://github.com/armtash/crits/blob/api_patch/crits/core/handlers.py#L660.  This means that a user could potentially add multiple source instances with the same date/key, which might cause issues when it comes time to modify/delete the source instance.

### DIFF
--- a/crits/core/api.py
+++ b/crits/core/api.py
@@ -601,7 +601,14 @@ class CRITsAPIResource(MongoEngineResource):
             'Email': {},
             'Event': {},
             'Exploit': {},
-            'Indicator': {},
+            'Indicator': {
+                'set_indicator_attack_type' : indh.set_indicator_attack_type,
+                'set_indicator_threat_type' : indh.set_indicator_threat_type,
+                'activity_add' : indh.activity_add,
+                'activity_update' : indh.activity_update,
+                'activity_remove' : indh.activity_remove,
+                'ci_update' : indh.ci_update
+                          },
             'IP': {},
             'PCAP': {},
             'RawData': {},
--- a/crits/core/api.py
+++ b/crits/core/api.py
@@ -567,6 +567,7 @@ class CRITsAPIResource(MongoEngineResource):
         import crits.objects.handlers as objh
         import crits.relationships.handlers as relh
         import crits.services.handlers as servh
+        import crits.indicators.handlers as indh
 
         actions = {
             'Common': {
@@ -574,6 +575,16 @@ class CRITsAPIResource(MongoEngineResource):
                 'add_releasability': coreh.add_releasability,
                 'forge_relationship': relh.forge_relationship,
                 'run_service': servh.run_service,
+                'status_update' : coreh.status_update,
+                'ticket_add' : coreh.ticket_add,
+                'ticket_update' : coreh.ticket_update,
+                'ticket_remove' : coreh.ticket_remove,
+                'source_add_update': coreh.source_add_update,
+                'source_remove': coreh.source_remove,
+                'action_add' : coreh.action_add,
+                'action_update' : coreh.action_update,
+                'action_remove' : coreh.action_remove,
+                'description_update' : coreh.description_update,
             },
             'Actor': {
                 'update_actor_tags': ah.update_actor_tags,
--- a/crits/core/handlers.py
+++ b/crits/core/handlers.py
@@ -61,56 +61,57 @@ from crits.core.totp import valid_totp
 
 logger = logging.getLogger(__name__)
 
-def action_add(obj_type, obj_id, action):
+def action_add(type_, id_, tlo_action, user=None, **kwargs):
     """
     Add an action to a TLO.
 
-    :param obj_type: The class type of the top level object.
-    :type obj_id: str
-    :param obj_id: The ObjectId of the to level object to update.
-    :type obj_id: str
-    :param action: The information about the action.
-    :type action: dict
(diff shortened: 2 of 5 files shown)


armtash:
Here are some examples.  I can post the other examples as well.

Add an action to a TLO:

         import requests
         url = 'http://localhost:8000/api/v1/<TLO_TYPE>/<TLO_ID>'
		 headers = {
                'Content-Type': 'application/json',
				}
         params = {
                      'api_key' : '<your api key>',
                      'username' : '<your username>',
                  }
         new_action = {
					'action_type' : '<action_type>',
					'active' : '<True OR False>',
					'begin_date' : '<begin_date>',
					'end_date' : '<end_date'>,
					'performed_date' : '<performed_date>'
					'reason' : '<reason for the action>'
					'date' :  '<action added date>'
					}
         data =  {
                      'action': 'action_add',
                      'tlo_action' : new_action,
                      }
         r = requests.patch(url, data=data, params=params, headers=headers )
         print r.content
		
Update an existing action:

         import requests
         url = 'http://localhost:8000/api/v1/<TLO_TYPE>/<TLO_ID>'
		 headers = {
                'Content-Type': 'application/json',
				}
         params = {
                      'api_key' : '<your api key>',
                      'username' : '<your username>',
                  }
         new_action = {
					'action_type' : '<action_type>',
					'active' : '<True OR False>',
					'begin_date' : '<begin_date>',
					'end_date' : '<end_date'>,
					'performed_date' : '<performed_date>'
					'reason' : '<reason for the action>'
					'date' :  '<date of action being modified>'
					}
         data =  {
                      'action': 'action_update',
                      'tlo_action' : new_action,
                      }
         r = requests.patch(url, data=data, params=params, headers=headers )
         print r.content


Remove an action from a TLO:

         import requests
         url = 'http://localhost:8000/api/v1/<TLO_TYPE>/<TLO_ID>'
         params = {
                     'api_key' : '<your api key>',
                      'username' : '<your username>',
                  }
         data =  {
                      'action': 'action_remove',
                      'date' : '<date of action to be removed>',
                      }
         r = requests.patch(url, params=params, data=data)
         print r.content
		
Update the description of a TLO:

         import requests
         url = 'http://localhost:8000/api/v1/<TLO_TYPE>/<TLO_ID>'
         params = {
                     'api_key' : '<your api key>',
                      'username' : '<your username>',
                  }
         data =  {
                      'action': 'description_update',
                      'description' : '<description>',
                      }
         r = requests.patch(url, params=params, data=data)
         print r.content

Set the attack type of an Indicator object:

         import requests
         url = 'http://localhost:8000/api/v1/indicators/<INDICATOR_ID>'
         params = {
                     'api_key' : '<your api key>',
                     'username' : '<your username>',
                  }
         data =  {
                      'action': 'set_indicator_attack_type',
                      'attack_type' : '<attack_type>',
                      }
         r = requests.patch(url, params=params, data=data)
         print r.content

mgoffin:
Awesome! I'll have to give this all a try when I get a chance. It looks great though and I agree with basically everything you said above :)

PR #665, opened 2015-12-03
When adding actions to a TLO, validate that the action type can be added

thelok:
When adding actions to a TLO, validate that the action type can be added and also the entire form itself.

`form.is_valid()` was being called but no action was taken on it.... so a user could add any action type they wanted? Is that the desired behavior?

Refactored `get_actions_for_tlo()` and created `get_action_types_for_tlo()` in handlers so the code can be reused.


mgoffin:
`is_valid` was being called because internally (to Django) it is what creates the `cleaned_data` out of the form which was used on the next line.

Adding the validation is good, though!

PR #672, opened 2015-12-16
Identify the newest available template for each service on get_sample_details

ms-xy:
Solution: This pull request will identify the newest available template for a service.

Issue: CRITs queries the analytic_results database to identify the a template path of a service. This in theory could allow CRITs to use a template that is unique for that version of the service. However, the way CRITs is used does not allow this to occur. This is because the services overwrite their core code on every new version. Additionally, the previous method would make it cumbersome to add or change custom templates.

### DIFF
--- a/crits/samples/handlers.py
+++ b/crits/samples/handlers.py
@@ -169,6 +169,18 @@ def get_sample_details(sample_md5, analyst, format_=None):
 
         # analysis results
         service_results = sample.get_analysis_results()
+        
+        # template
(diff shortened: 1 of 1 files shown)


webstergd:
For additional background:
We are currently using this to help with distributed services. Specifically, this is being used to change the display engine without having to manually manipulate the database for each analytic result. With respect to Distributed Cuckoo, this has allowed us to store the api call traces in the database without having to clutter the web gui.

PR #674, opened 2015-12-21
Added support for Indicator description field when creating new Indicators

frbapolkosnik:
Submissions through UI, Blob, and CSV should work now.
I've not tested the API, but it should work as well.


mgoffin:
You'll want to update help_main.txt with the new "description" API parameter for Indicators and run the `rst2html.py` file to generate the help_main.html file with your changes.

frbapolkosnik:
lol, I can't type today.

PR #675, opened 2015-12-30
Update Signature TLO, Take Advantage of the PATCH API

dreardon:
Continuing with the PATCH work started in #449 and further continued in #664, I have made an update to the Signature handlers to conform with requests using the PATCH API. Through this feature, the following Signature-specific requests can now be made:
 - update_dependency
 - update_min_version
 - update_max_version
 - update_signature_data
 - update_signature_type
 - update_title


mgoffin:
Nice :D

PR #693, opened 2016-02-08
Django 1.8 fixes

frbapolkosnik:
It should be working now. Just need to remove some comments.

### DIFF
--- a/crits/core/views.py
+++ b/crits/core/views.py
@@ -346,7 +346,7 @@ def login(request):
     user_agent = request.META.get('HTTP_USER_AGENT', '')
     remote_addr = request.META.get('REMOTE_ADDR', '')
     accept_language = request.META.get('HTTP_ACCEPT_LANGUAGE', '')
-    #next_url = request.REQUEST.get('next', None)
+    next_url = request.REQUEST.get('next', None)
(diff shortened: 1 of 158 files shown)


crits/domains/handlers.py
```diff
@@ -2,6 +2,7 @@
 import re
 import datetime
 
+from pymongo import read_preferences
```
mgoffin: What's this for?

crits/core/views.py
```diff
@@ -346,7 +346,7 @@ def login(request):
     user_agent = request.META.get('HTTP_USER_AGENT', '')
     remote_addr = request.META.get('REMOTE_ADDR', '')
     accept_language = request.META.get('HTTP_ACCEPT_LANGUAGE', '')
-    next_url = request.REQUEST.get('next', None)
+    #next_url = request.REQUEST.get('next', None)
```
mgoffin: How do we go to the URL the person was trying to go to without this?

crits/domains/handlers.py
frbapolkosnik: I think that was just an incidental paste/left over from some experimentation.

PR #709, opened 2016-03-03
Make paths relative

tsmolka:
Deploying CRITs in a subdirectory is currently quite challenging. Static files are prefixed with {{STATIC_URL}} already, but the "media files" (by default from extras/www) are assumed to be placed in web server root as well. 

By doing few simple changes (adding <base> meta tag to html documents and making paths relative) I was able to deploy CRITs in a subdirectory. Since MEDIA_URL is by default "/" I believe these changes should be safe. There might still be some specific instances of absolute paths I was not able to identify. This would however impact only users like me who change default MEDIA_URL.

### DIFF
--- a/crits/campaigns/templates/campaign_monthly.html
+++ b/crits/campaigns/templates/campaign_monthly.html
@@ -5,7 +5,7 @@
 
 {% block content %}
 <h2>Email chart for {{campaign}}</h2>
-   <script src="/js/flot/jquery.flot.js" type="text/javascript"></script>
+   <script src="js/flot/jquery.flot.js" type="text/javascript"></script>
    <script>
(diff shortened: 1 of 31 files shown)


tsmolka:
Added commit 0f12f18 to solve [https://www.tjvantoll.com/2013/02/17/using-jquery-ui-tabs-with-the-base-tag/](url)

mgoffin:
I'm confused by this. The web server root is configurable in your apache config and is basically a one-line change.

tsmolka:
I am talking about situation when you mount Django app on subpath. You want it to be accessible via prefix (e.g. https://private-server/crits/) and not directly via base url (e.g. https://private-server/).

PR #710, opened 2016-03-08
Update versions in DEPENDENCIES to match those in requirements.txt

brlogan:
This simply updates conflicting version numbers to reduce confusion. I did not try to add anything new to DEPENDENCIES from requirements.txt.


brlogan:
Can we pull this in, or do I need to change something?

PR #716, opened 2016-03-11
mongoengine 0.9.0 support added

frbapolkosnik:
Update to use mongoengine0.9.0

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -41,17 +41,17 @@ Provides: PIL
 PyYAML
 Provides: yaml
 
-django<1.7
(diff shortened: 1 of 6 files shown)


frbapolkosnik:
I think that it is good to go.

brlogan:
Should probably update version in DEPENDENCIES too.

mgoffin:
Yea, I just pulled that PR in (thanks for the reminder). DEPENDENCIES and requirements should be kept in-line.

frbapolkosnik:
Updated DEPENDENCIES, good to go.

PR #718, opened 2016-03-14
Add related dropdown 

robertsjw:
This adds a dropdown on the details view of Top Level Objects, allowing for users to quickly add another related object. 

I welcome any testing / feedback :)


frbapolkosnik:
I've noticed that you have a few spots whete you have:
`self.fields['relationship_type'].initial = "Related To"`
you need to use Vocab there:
`from crits.vocabulary.relationships import RelationshipTypes
...
self.fields['relationship_type'].initial = RelationshipTypes.RELATED_TO`

robertsjw:
That makes sense, done.

mgoffin:
Thank you so much for adding this!

PR #722, opened 2016-03-19
Support for Mongoengine 0.10.x, and Pymongo 3.2.2

frbapolkosnik:
This PR facilitates backwards compatibility with mongoengine 0.8.8 and Django 1.8.x, while allowing you to use currently supported version of mongoengine (0.10.x), so in essence everything should continue to work the way it always has without any changes.

If you want to go with mongoengine 0.10.x, you'll need to upgrade via pip.
```pip install -U -r requirements.txt```

It will install Mongoengine 0.10.6 and Pymongo 3.2.2 and then also django_mongoengine from git.

~~This needs some more testing. In my limited testing, so far I've noticed:
My older database contained 'rar_path', and that used to throw an exception. I've added "self._meta['strict'] = False" in the __init__ of CritsDocument, and mongoengine 0.9+ becomes more forgiving about those old fields.~~

**CAVEATS:** there are two bleeding edge packages being used (from git), see requirements.txt
It works with Django 1.8, if you want to get it to work on Django 1.9 see a note below.
(That is until they merge the Django 1.9 fix)

**Possible Problems:**
If you want to get it to work on Django 1.9... take a look here: https://github.com/wlanslovenija/django-tastypie-mongoengine/pull/94 

The official django-tastypie-mongoengine Github repo (https://github.com/wlanslovenija/django-tastypie-mongoengine) is unmaintained. If there's somebody that can invest some spare cycles to merge some of the PRs to enable future compatibility with more recent versions of Django and mongoengine. Or somebody could just rewrite the whole REST API to use a different framework like DRF.

In meantime, I've noticed that https://github.com/wishtack/django-tastypie-mongoengine seems to be active, and tagged two new releases, so I've pointed the git to use this repo, since it has the fix to support mongoengine 0.10.x. I also have a branch at https://github.com/frbapolkosnik/django-tastypie-mongoengine that has also some fixes for Python 3.x compatibility, but I can't guarantee how long it's going to be around.

Enjoy, and let me know if you run into any issues if you decide to test this.

Thanks,
-A


mgoffin:
You note that tastypie-mongoengine is broken with newer versions. Does that mean the API is still broken?

frbapolkosnik:
Using the git version of django-tastypie-mongoengine fixes that.

frbapolkosnik:
Though, I've not tested it yet.

apolkosnik:
I've testeddomain addition and sample additions via the API, both worked. Great success! ;-)

apolkosnik:
It looks like no matter what I do( drop the db, checkout master, etc), those 3 tests fail on my test machine.

frbapolkosnik:
Ok, these test errors were specific to the recent changes in relationship code, PR #769 addresses those.

frbapolkosnik:
I think that this is good to go. Once the admin upgrades via pip to the new mongoengine/pymongo some changes are happening, but without upgrading everything should work just fine.

mgoffin:
I'm going to accept this, but please be on the lookout for people having installation issues. I've only been able to test so much and there are probably a lot of people on different systems which might have issues we need to address immediately!

brlogan:
Did I miss something, or does this require installing "future" so that the `import builtins` line of django-tastypie-mongoengine works?

apolkosnik:
I must have overlooked that

apolkosnik:
added the future to the setup.py

Issue #720, opened 2016-03-17
Login SESSION_KEY:  value must be an integer


thelok:
Installed Django 1.8.11. Trying to login to CRITs with a remote user yields the following 500 error related to SESSION_KEYs:

```
[Thu Mar 17 15:14:50 2016] [error] [client ...]   File ".../lib/python2.7/site-packages/django/contrib/auth/__init__.py", line 59, in _get_user_session_key
[Thu Mar 17 15:14:50 2016] [error] [client ...]     return get_user_model()._meta.pk.to_python(request.session[SESSION_KEY])
[Thu Mar 17 15:14:50 2016] [error] [client ...]   File ".../lib/python2.7/site-packages/django/db/models/fields/__init__.py", line 969, in to_python
[Thu Mar 17 15:14:50 2016] [error] [client ...]     params={'value': value},
[Thu Mar 17 15:14:50 2016] [error] [client ...] ValidationError: [u"'521238e4550671234123c1' value must be an integer."]
```

I see a bunch of comments in core/user.py and core/handlers.py and some changes in settings.py with the middleware stuff in addition to a custom middleware class in user.py.

Tried making a bunch of permutations and fiddling with the MIDDLEWARE_CLASSES but nothing seems to work.

Is there no workaround right now because of mongoengine?

https://github.com/crits/crits/blob/master/crits/core/handlers.py#L3363

frbapolkosnik:
Are you on master?

mgoffin:
And using mongoengine 0.8.8?

thelok:
Tried both master and stable_4 with 0.8.8 on both branches.

frbapolkosnik:
Take a look there:
https://github.com/crits/crits/blob/master/crits/core/handlers.py#L3369

frbapolkosnik:
We override the django piece, because mongoengine makes us do so.

thelok:
Holy crap I think I just found it:

--- a/crits/settings.py
+++ b/crits/settings.py
@@ -474,9 +474,10 @@ if REMOTE_USER:
         'django.middleware.common.CommonMiddleware',
         'django.contrib.sessions.middleware.SessionMiddleware',
         'django.contrib.auth.middleware.AuthenticationMiddleware',
-        'django.contrib.auth.middleware.RemoteUserMiddleware',
+        'django.middleware.csrf.CsrfViewMiddleware',
         'django.contrib.messages.middleware.MessageMiddleware',
[...]

PR #723, opened 2016-03-20
fix SESSION_KEY for REMOTE_USER

razuz:
Fixes issue #720


apolkosnik:
It looks good to me

mgoffin:
I'll check this out before accepting. I am wary about what might break with the CRSF middleware in there. Specifically forms and services.

thelok:
So since the CSRF middleware wasn't in there for people that used REMOTE_USER, what is the implication of that?

PR #727, opened 2016-03-23
Sort services case insensitively, sort users by login date by default.

thelok:
This has been annoying me for the LONGEST time. These are control panel changes.

Changed users page to sort by last login date by default.

Changed services page to sort by name, case insensitively.


thelok:
We can do without the user sort by login date in this pull request if you want but I'm keeping it in my instance.

PR #729, opened 2016-03-24
Fixes some issues with dialogs.

mgoffin:
core.js was missing in a collectstatic run so it was caught here. Some
of the dialogs which had a function mapped when opening lost it because
it was consumed by the related newPersona. This adds them back.

Also fixed an issue with uploading Samples using 7zip where the
appropriate relationship fields were being passed in but the arguments
weren't there so it was failing.


robertsjw:
:+1:  Nice catch, thank you :)

Issue #616, opened 2015-10-06
Events API doesn't allow campaign or confidence


Magicked:
According to the docs at https://github.com/crits/crits/wiki/Authenticated-API, most TLOs should allow campaign and confidence assignment (I realize it says "most"). When creating events through the API, it doesn't look like campaign or confidence is used. Is this something we should add or was there a reason for not doing this?

mgoffin:
This is mainly because the handler for creating Events doesn't allow for it:

https://github.com/crits/crits/blob/master/crits/events/handlers.py#L229

PR #741, opened 2016-04-15
Added Campaign assignment to Events

Magicked:
This is a proposed fix for #616 

Campaign and confidence assignment for Event objects can now be accomplished through both the API and the "New Event" form in the GUI. I've tested this change in several different cases (with campaign, without campaign, API, etc), and I have had no issues.

Let me know if anything needs to be tweaked, and if so I will resubmit the PR.


mgoffin:
Seems to work on my end! If there's any corner-cases found we can fix them up in master. Thanks!!

PR #744, opened 2016-04-29
Improvements to upgrade script and migrations

brlogan:
This smooths out the upgrade progress with better error handling and more information to help solve migration issues.

### DIFF
--- a/crits/core/core_migrate.py
+++ b/crits/core/core_migrate.py
@@ -18,7 +18,8 @@ def migrate_analysis_results(self):
                 del result['id']
             config = result.get('config', {})
             ar.config = AnalysisConfig(**config)
-            del result['config']
+            if 'config' in result:
+                del result['config']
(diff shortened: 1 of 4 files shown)


mgoffin:
If I am reading this correctly this change will blow up your shell with output instead of prior where it would update a single line with the new count. This was beneficial not only from keeping output to a minimum but also giving realtime readable updates without needing massive scroll back to see what's been going on. Any particular reason for changing that?

brlogan:
Oops, just realized that I never responded to this.

This code still updates a single line in the shell with the count because it still contains the "\r" carriage return control character.
Shell output is actually reduced a bit because it no longer lists every single backdoor and exploit ID that is being migrated.  The only time you should get additional shell output is in the form of extra information to help solve TLO migration errors.

mgoffin:
Ok. Looks like this needs an update with master before it can get merged :)

PR #747, opened 2016-05-05
Remove patterns() for Django future compatibility 

frbapolkosnik:
Deprecated since version 1.8: urlpatterns should be a plain list of django.conf.urls.url() instances instead.

https://docs.djangoproject.com/en/1.8/ref/urls/#django.conf.urls.patterns


frbapolkosnik:
Technically, I went a bit too far on this one, all was needed was to wrap the paths in url(), but I replaced patterns(), and added the specific prefix= to every url() call. Everything is still good. I guess we might

PR #752, opened 2016-05-17
Modify type/value pair extraction for new vocab

brlogan:
When loading a TLO details page that has Objects, the code attempts to determine if there are any matching Indicators by analyzing the TLO's relationships list. The code responsible for this was not adjusted to work with the new vocabulary and thus looks for the old "name" field that no longer exists.


mgoffin:
Looks good but needs a merge with master evidently.

PR #753, opened 2016-06-03
Improve Performance of Object & Relationship Creation

brlogan:
Bulk uploading large quantities of Objects was taking an extremely long time, especially when the "Add Indicator" option was selected. I found that a huge portion of the time was spent saving the entire TLO back to the DB after adding the Object, and then again after adding the relationship to the Indicator.  As more and more Objects/Relationships are added, the document takes even longer to write to the DB. By using `TLO.update(add_to_set__xxxx=xxxx)` to insert the new object/relationship, rather than `TLO.save()`, uploading a sample set of 100 objects (and adding Indicators for each) took ~10% as long to complete. 

I think there may be additional performance improvements by caching the database writes until each block of 50 objects have been processed, but the solution in this PR is a good start and also noticeably improves performance when adding a single relationship or object to any TLO with a very large number of relationships or objects.

### DIFF
--- a/crits/core/crits_mongoengine.py
+++ b/crits/core/crits_mongoengine.py
@@ -1686,7 +1686,7 @@ class CritsBaseAttributes(CritsDocument, CritsBaseDocument,
 
     def add_relationship(self, rel_item, rel_type, rel_date=None,
                          analyst=None, rel_confidence='unknown',
-                         rel_reason='N/A', get_rels=False):
+                         rel_reason='', get_rels=False):
         """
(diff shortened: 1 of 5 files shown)


cdorer:
Nice find!

frbapolkosnik:
Nice!

mgoffin:
Seems to work well for me! I tested with:

* single Object addition
* bulk Object addition
* Indicator creation with Object addition
* Relationship addition

Thanks for speeding things up! I'm going to accept to master and if any bugs surface we can fix.

PR #755, opened 2016-06-04
Allow triage of Samples with no binary

brlogan:
Binary data shouldn't be required for triage to be run on a Sample. Valuable analysis can be performed using only the Sample's metadata.  This change was originally submitted as PR #1375 of legacy, but was lost during the open source transition.


mgoffin:
Could you update this with master and we can merge it in?

brlogan:
GitHub is telling me that this has no conflicts with master.  Is it telling you something different?

mgoffin:
Yea I think iOctocat was having issues as it wouldn't let me merge it.

PR #756, opened 2016-06-04
Check for "seek" attribute before seeking

brlogan:
Occasionally `local_obj.obj.filedata` will evaluate to True, but it won't actually have the `.seek()` attribute. This prevents an exception in that case.


mgoffin:
Can you update with master and we can merge this in?

PR #757, opened 2016-06-06
Fix #677

ManoftheSea:
Load of an object from mongo with partial data would cause a schema update on partial data.  This is bad.  It is important to avoid an infinite loop on reload.

Issue #677, opened 2016-01-08
Cannot open details on IPs since upgrade


cjdupreez:
I recently upgraded my CRITs instance to the latest version via git pull and now it seems I can't open a large portion of my previously submitted IPs. When I click on details, I get the following response:

Request Method: GET
Request URL: https://x.x.x.x/ips/details/85.128.142.31/

Django Version: 1.6.1
Python Version: 2.7.6
Installed Applications:
('crits.core',
 'crits.dashboards',
 'django.contrib.auth',
 'mongoengine.django.mongo_auth',
 'django.contrib.contenttypes',
 'django.contrib.messages',
 'django.contrib.sessions',
 'django.contrib.sites',
 'django.contrib.staticfiles',
 'crits.actors',
 'crits.campaigns',
 'crits.certificates',
 'crits.domains',
 'crits.emails',
 'crits.events',
 'crits.indicators',
 'crits.ips',
 'crits.locations',
 'crits.objects',
 'crits.pcaps',
 'crits.raw_data',
 'crits.relationships',
 'crits.samples',
 'crits.screenshots',
 'crits.services',
 'crits.signatures',
 'crits.stats',
 'crits.targets',
 'tastypie',
 'tastypie_mongoengine')
Installed Middleware:
('django.middleware.common.CommonMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware')

Traceback:
File "/usr/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  112.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/lib/python2.7/dist-packages/django/contrib/auth/decorators.py" in _wrapped_view
  22.                 return view_func(request, *args, **kwargs)
File "/data/crits/crits/ips/views.py" in ip_detail
  68.                                           analyst)
File "/data/crits/crits/ips/handlers.py" in get_ip_details
  163.         ip.sanitize("%s" % analyst)
File "/data/crits/crits/core/crits_mongoengine.py" in sanitize
  2441.                     self.sanitize_relationships(username, sources)
File "/data/crits/crits/core/crits_mongoengine.py" in sanitize_relationships
[...]

PR #760, opened 2016-06-08
bump in Indicator schema version requires a run of the update script.

ManoftheSea:


mgoffin:
This is not required, only optional. The whole point of the "auto-migrate on access" is so people can restart the UI after pulling code in and data can migrate as it is used. People can optionally run the upgrade script to push things along even after the UI is restarted.

ManoftheSea:
As fix for #677 (#757) was accepted, this log update really is optional.

PR #763, opened 2016-06-11
default exclude param to empty list

ckane:
Otherwise, line 171 may try to treat None as a list and crash. This has been breaking sample API searches for me.

Can someone else give this a whirl for me, just to verify that it doesn't break a thing?

### DIFF
--- a/crits/samples/sample.py
+++ b/crits/samples/sample.py
@@ -158,7 +158,7 @@ class Sample(CritsBaseAttributes, CritsSourceDocument, CritsActionsDocument,
         if isinstance(filenames, list):
             self.filenames = filenames
 
-    def _json_yaml_convert(self, exclude=None):
+    def _json_yaml_convert(self, exclude=[]):
         """
         Helper to convert to a dict before converting to JSON.
 


ckane:
@mgoffin @wxsBSD

apolkosnik:
I'll take it for a whirl later tonight

ckane:
Sweet, thanks!

apolkosnik:
Looks good, if the exclude is set to None an exception is thrown "TypeError: 'NoneType'  is not iterable", your fix fixes that case. Also,  if there is no exclude defined, we're setting the default to None instead of a list, as it is stated in the comments. It looks like we might need to make a few fixes in crits_mongoengine.py

apolkosnik:
Looks good, if the exclude is set to None an exception is thrown "TypeError: 'NoneType'  is not iterable", your fix fixes that case. Also,  if there is no exclude defined, we're setting the default to None instead of a list, as it is stated in the comments. It looks like we might need to make a few fixes in crits_mongoengine.py

ckane:
Ok, you're thinking the following from <code>crits_mongoengine.py</code>?:
* <code>CritsQuerySet.to_json</code>
* <code>CritsQuerySet.to_yaml</code>
* <code>CritsDocument._json_yaml_convert</code>
* <code>CritsDocument.to_json</code>
* <code>CritsDocument.to_yaml</code>

I'll go ahead and merge this one in and then start a new branch/PR for those (plus more found).

PR #766, opened 2016-06-15
Sample Relationships To TLO Overwritten By Triage Object

thelok:
Fix an issue where data would be overwritten by a triage.

This change just moves the relationships and saving of those relationships to BEFORE the triage occurs.

### DIFF
--- a/crits/samples/handlers.py
+++ b/crits/samples/handlers.py
@@ -814,12 +814,12 @@ def handle_file(filename, data, source, method='Generic', reference='',
             sample.filenames.append(filename)
         if cached_results != None:
             cached_results[md5_digest] = sample
-    
+
     if not sample.description:
         sample.description = description
     elif sample.description != description:
         sample.description += "\n" + description
-   
+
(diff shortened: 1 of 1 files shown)


thelok:
Lastly I'll say this probably affected samples so that they have unidirectional relationships instead of bidirectional. (e.g. Email -> Sample but not Sample -> Email)

mgoffin:
As in it fixed that problem, or that with this being accepted samples will now have unidirectional relationships?

thelok:
Future samples will have the correct bidirectional relationship. But existing samples still have this issue. Nothing has been retroactively fixed.

thelok:
On our end we saw it when we upload an EML with attachments and the the macro_extract service created RawData objects. The Sample->RawData relationship exists, but Sample->Email did not.

You can reproduce this issue by uploading a "Related" sample where the sample will generate artifacts and other relationships (e.g. pdf2txt, macro_extract), the sample should not have a relationship back to the original TLO it was related to.

PR #770, opened 2016-06-30
Remove reference to missing theme image file

brlogan:
One of the image files for the "Fil" query-theme is missing: ui-bg_highlight-soft_75_d7d5c0_1x100.png
Changing the "background" property to just use the color.

### DIFF
--- a/extras/www/css/jquery-themes/Fil/style.css
+++ b/extras/www/css/jquery-themes/Fil/style.css
@@ -17,7 +17,7 @@ div.top {
     margin: 0px;
     vertical-align: middle;
     height: 25px;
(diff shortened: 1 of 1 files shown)


apolkosnik:
Looks good

mgoffin:
All of the other themes use this:

`background: url('/new_images/top_bar_bg.png') repeat-x;`

Just curious why this isn't the fix here?

brlogan:
I figured that Fil intentionally chose a different color than the one that is provided by top_bar_bg.png

mgoffin:
I just tried changing to the Fil theme and even after clearing browser cache it didn't look very different, haha. Does adding the line above make things look weird or still fine?

brlogan:
It makes the top and bottom bars a tan color for me.

mgoffin:
I guess it's fine then. If there's an issue at least we can look back at this and see what the other themes do.

PR #773, opened 2016-07-01
impfuzzy hashes

frbapolkosnik:
This is just a discussion piece at this point, as probably it could be implemented without actually reliance on pyimpfuzzy. (https://github.com/JPCERTCC/aa-tools/tree/master/impfuzzy/pyimpfuzzy)
In any event, pyimpfuzzy in this patch is clearly optional.

I'm not sure how this should work, but I figured that it makes sense to store impfuzzy values in Sample together with other hashes, so that we can use them similarly to ssdeep.

With that one can create a service based on ssdeep_compare to find PE files with varied similarity of imports.


mgoffin:
I think this makes sense! Would probably be worth adding a migrate which only builds and saves the hash if a backing binary exists. Then people could run 'upgrade -sS' to build their hashes up.

mgoffin:
That migrate isn't going to work for a few reasons:

1) we never want to always migrate every time a sample is pulled from the database (this will happen, even if it does nothing)
2) pyimpfuzzy isn't imported to actually do the work here.

The schema should be bumped to 5, and a `migrate_4_to_5()` should be made.

crits/samples/migrate.py
```diff
@@ -72,7 +72,40 @@ def migrate_sample(self):
     Migrate to the latest schema version.
     """
 
-    migrate_3_to_4(self)
+    migrate_4_to_5()
```
mgoffin: Should pass `self` into this.

crits/samples/migrate.py
```diff
@@ -72,7 +72,40 @@ def migrate_sample(self):
...
+            #in case we had some Pes that were not covered
+            try:
+               self.impfuzzy = pyimpfuzzy.get_impfuzzy_data(self.filedata.read())
+            except Exception:
+                # The sample is probably not a PE wit valid imports or we don't have pyimpfuzzy
+                self.impfuzzy = None
+    else:
+        # not a PE, so no point in populating it
+        self.impfuzzy = None
+        self.schema_version = 5
```
mgoffin: Putting this in the else clause prevents proper schema bumping for things that don't match the else, and they also don't get saved or reloaded.

mgoffin:
I think that looks good. I'll give it a go. There's no way to install pyimpfuzzy besides getting it from that site? Or does `pip install` work?

apolkosnik:
Unfortunately, no pip package, and the pip install via git is not going to work either, since they have it one level too high.

crits/samples/migrate.py
```diff
@@ -72,7 +72,41 @@ def migrate_sample(self):
...
+            import pyimpfuzzy
+        except ImportError:
+            pass
+        if not hasattr(self, 'impfuzzy'):
+            try:
+                self.impfuzzy = pyimpfuzzy.get_impfuzzy_data(self.filedata.read())
+            except Exception:
+                # The sample is probably not a PE with valid imports or we don't have pyimpfuzzy
+                self.impfuzzy = None
+        elif self.impfuzzy None:
```
mgoffin: Syntax error

mgoffin:
Should we update the dependencies with pyimpfuzzy now that you can pip install it?

frbapolkosnik:
"add pyimpfuzzy to dependencies" commit is actually an update to requirements.txt

crits/samples/migrate.py
```diff
@@ -72,7 +72,31 @@ def migrate_sample(self):
...
+    if self.schema_version < 4:
+        migrate_3_to_4(self)
+
+    if self.is_pe():
+        try:
+            import pyimpfuzzy
+            if not self.impfuzzy:
+                self.impfuzzy = pyimpfuzzy.get_impfuzzy_data(self.filedata.read())
+        except Exception:
+                print("\n\tError migrating %s: %s" % (self.id, "The sample is probably not a PE with valid imports or we don't have pyimpfuzzy"))
```
mgoffin: This needs to be removed.

mgoffin:
I think this is ready to go? Anything else you're working on? I ran the migrate and everything that I expected to have the hash now has it.

frbapolkosnik:
Good to go.

PR #777, opened 2016-07-02
Handle Source-less TLOs for Comment notifications

brlogan:
When a user attempts to add a Comment to a TLO that doesn't have a Source field, they are given no indication that the comment was added successfully. No response message is displayed and the page is not updated. However, if the page is reloaded, the user can see that the comment was indeed added.
This is because adding the comment triggers the creation of a notification. The code attempts to grab the source from the TLO, but since the TLO has no Source, an error occurs. The code now only pulls source data if the object has source data.

### DIFF
--- a/crits/notifications/handlers.py
+++ b/crits/notifications/handlers.py
@@ -41,6 +41,7 @@ def create_notification(obj, username, message, source_filter=None,
     n = Notification()
     n.analyst = username
     obj_type = obj._meta['crits_type']
+    users = set()
 
     if notification_type not in NotificationType.ALL:
         notification_type = NotificationType.ALERT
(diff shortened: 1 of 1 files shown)


mgoffin:
Just making sure, but does this allow users to add comments to TLOs they do not have source access to?

brlogan:
The modified function only handles creation of notifications and it only uses sources to determine if a user who is subscribed to the given TLO should actually receive a notification about it.  If the TLO has sources to check against, then it makes sure each subscribed user can view one of those sources prior to adding them to the list of recipients.  If the TLO does not have sources to check against, any subscribed user will get a notification about it.
Someone could mention a user in a comment on a TLO that the mentioned user doesn't have access to, and that person would receive the notification, but they of course would still not be able to view the TLO.

This change doesnt' affect who can actually add comments to TLOs.

PR #780, opened 2016-07-05
Fix a couple issues with the Event Count feature

brlogan:


brlogan:
This is in response to comment by @r3comp1le on 15bd717edb3b1dce37751c5302a953fdfa81d35e.

PR #782, opened 2016-07-12
Refactor, cleanup, and fix Email functions

brlogan:
The EML handle functions still had parent id/type parameters, and the Add Sample button on Events used an upload_sample function, that had both been effectively replaced by the general Add Related TLO functionality, so it made sense to remove them.  In the process I fixed an issue with the Indicator Blob/CSV upload, and made the email functions more consistent with each other.  While I was at it, I tried to make things respect the PEP8 79 character limit.

### DIFF
--- a/crits/emails/handlers.py
+++ b/crits/emails/handlers.py
@@ -25,7 +25,7 @@ from django.template.loader import render_to_string
 
 from crits.campaigns.forms import CampaignForm
 from crits.config.config import CRITsConfig
-from crits.core.crits_mongoengine import json_handler, create_embedded_source
+from crits.core.crits_mongoengine import json_handler
 from crits.core.crits_mongoengine import EmbeddedCampaign
 from crits.core.data_tools import clean_dict
 from crits.core.exceptions import ZipFileError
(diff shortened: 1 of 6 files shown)


crits/emails/handlers.py
```diff
@@ -548,10 +545,6 @@ def handle_email_fields(data, analyst, method, related_id=None, related_type=Non
                               date=datetime.datetime.now())
         new_email.add_campaign(ec)
 
-
-    new_email.save(username=analyst)
```
mgoffin: With this removed, when does it get saved to the database to get an ObjectId assigned so it can be used when forming any relationships necessary?

crits/emails/handlers.py
```diff
@@ -560,12 +553,10 @@ def handle_email_fields(data, analyst, method, related_id=None, related_type=Non
...
-                                          relationship_type,
-                                          analyst=analyst,
-                                          get_rels=False)
+
+    new_email.add_source(source=sourcename, method=method,
+                         reference=reference, analyst=analyst)
+
 
     try:
         new_email.save(username=analyst)
```
brlogan: The email gets saved here after the Source and Campaign get added.  Relationships aren't formed until later.

mgoffin:
Ahh, that wasn't showing in my diff. That would do it!

Issue #784, opened 2016-07-13
API doesn't allow descrip in Samples


mike-bailey:
Cannot assign a description to samples via the [Authenticated API](https://github.com/crits/crits/wiki/Authenticated-API) unless I'm missing something. Technically an enhancement.

mike-bailey:
I modified this and fixed it so whenever @mgoffin says they'd actually like this as a feature I can PR it.

mgoffin:
yea that would be a great feature! throw up the PR and we'll take a look!

PR #785, opened 2016-07-13
Include description option in API for CRITS

mike-bailey:
May also wanna update the Wiki with the fix. I just implemented it how the rest of stuff is.

### DIFF
--- a/crits/samples/api.py
+++ b/crits/samples/api.py
@@ -73,7 +73,6 @@ class SampleResource(CRITsAPIResource):
             filename = None
 
         campaign = bundle.data.get('campaign', None)
-        description = bundle.data.get('description', None)
         confidence = bundle.data.get('confidence', None)
         source = bundle.data.get('source', None)
         method = bundle.data.get('method', "")
--- a/crits/samples/api.py
+++ b/crits/samples/api.py
@@ -60,7 +60,6 @@ class SampleResource(CRITsAPIResource):
         if type_ == 'metadata':
             filename = bundle.data.get('filename', None)
             md5 = bundle.data.get('md5', None)
-            description = bundle.data.get('description', None)
             password = None
             filedata = None
         elif type_ == 'file':
@@ -74,6 +73,7 @@ class SampleResource(CRITsAPIResource):
             filename = None
 
         campaign = bundle.data.get('campaign', None)
+        description = bundle.data.get('description', None)
         confidence = bundle.data.get('confidence', None)
         source = bundle.data.get('source', None)
         method = bundle.data.get('method', "")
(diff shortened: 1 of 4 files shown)


frbapolkosnik:
Looks good to me

mike-bailey:
Related to https://github.com/crits/crits/issues/784
I'm sure this could be done to other endpoints (for instance /ips), I just don't have time to do it to other ones and I needed the samples descrip ASAP for a project.

crits/samples/api.py
```diff
@@ -83,6 +84,7 @@ def obj_create(self, bundle, **kwargs):
         related_type = bundle.data.get('related_type', None)
         backdoor_name = bundle.data.get('backdoor_name', None)
         backdoor_version = bundle.data.get('backdoor_version', None)
+	description = bundle.data.get('description', None)
```
frbapolkosnik: Is the indent off here?

mike-bailey:
Crap sorry one second. Lemme check.

crits/samples/api.py
mike-bailey: Seems so. One moment.

mike-bailey:
Bear with me on the spacing issue, editor bug.

crits/samples/api.py
mike-bailey: @frbapolkosnik good now?

mgoffin:
This requires an update to:

https://github.com/crits/crits/blob/master/documentation/help_main.txt

and then `rst2html.py help_main.txt help_main.html` to update the API documentation.

mike-bailey:
Sure. Also, it currently only does it if it's metadata. This worked for our use case but isn't the case if it's public. I'll fix that.

mike-bailey:
Should be fixed.

crits/samples/api.py
```diff
@@ -83,6 +84,7 @@ def obj_create(self, bundle, **kwargs):
         related_type = bundle.data.get('related_type', None)
         backdoor_name = bundle.data.get('backdoor_name', None)
         backdoor_version = bundle.data.get('backdoor_version', None)
+        description = bundle.data.get('description', None)
```
mgoffin: Wouldn't this just be a duplicate of the above description you've added?

mike-bailey: It's moved from the past commit, not copied. Unless I messed up somehow. Before it was under metadata, but it should be able to assign a descrip regardless.

mike-bailey: Ah you're right

mike-bailey:
Sorry if this kinda shifted targets @mgoffin, but the same thing but with /api/v1/ips/ became a priority too so I went ahead and included it. Also fixed duplicate.

crits/samples/api.py
mike-bailey: Fixed this in the previous commits in addition to changing IP.

PR #791, opened 2016-07-28
Make Indicator Threat and Attack Types a tag list.

mgoffin:
These are functionally equivalent to sectors. You can modify the list as
you see fit through the UI and clicking on one will search for all other
Indicators with that type specified.

There is a schema change and an entry in UPDATING outlining the change,
the updating command to run if you wish, how to find Indicator
duplicates in your database for *manual* reconciliation, and the changes
to the PATCH API for developers.


brlogan:
Looks good, but a few questions/comments:
- Do we still want to have an "Unknown" type? It seems to me that an empty list is essentially the same as Unknown. It does provide the benefit of clicking it to see all other Indicators with type "Unknown", but is that enough of a reason to keep it?
- If I click the "X" to remove the last remaining type, it disappears from the list, but isn't removed from the DB; if I reload the page it reappears.
- There is a lot of margin around the Threat Types and Attack Types boxes. What do you think about reducing it to like 1px? Would save some vertical space.

mgoffin:
@brlogan The Unknown type is there for exactly the reason you noted. Without the value we couldn't do search and comparison. The reason the last remaining type stays is because you have to have a value there. When uploading an Indicator you have to provide a choice and this reflects that. I also agree that there's a lot of space around them. I suspect this is stemming from the tagit JS library and its associated CSS but after mucking with it for a bit I decided to put off dealing with that as the Sector box has the same issue and it's been there for quite a while.

brlogan:
Yeah, this has always bothered me about the sector box, but it's worse now that it's multiplied by 3.

I just poked at the margin thing and I see two options:
1. Add `style="margin: 1px;` to each tag list "ul" like this:
`<ul style="margin: 1px;" id="attack_type_list">`
2. If you want to use CSS, you could add some class like "taglist" to all 3 tag lists:
`<ul class="taglist" id="threat_type_list"`
`<ul class="taglist" id="attack_type_list">`
`<ul class="taglist" id="sector_list">`
Then in core.css do:
```css
ul.taglist {
    margin: 1px;
}
```

mgoffin:
@brlogan how's that? :)

brlogan:
That should do it!

I'm on the fence about the "Unknown" type. From one perspective, getting rid of "Unknown" would match Sectors better (we don't have an Unknown sector option), save space in the DB, and prevent situations where we have both Unknown and something else in the same list (someone initially adds an Indicator with Unknown, then later uploads the same Indicator with a different attack/threat type).  Adding a blank option to the dropdowns for Indicator upload would be easy.
On the other hand, it's nice to be able to click "Unknown" and find the other Indicators.  Maybe we could add some JS that would check if the item being removed is the last one and popup a message explaining that the list must contain at least one item? It's misleading to have the last item disappear from the list and then reappear when the page is reloaded.

mgoffin:
I'll look into doing two things:

1) If the only thing in the list is `Unknown` and something is being added, remove `Unknown` first.
2) If there is only one thing in the list, disable the "remove" feature on it until a second one has been added.

That seem good?

brlogan:
Seems reasonable to me! :)

mgoffin:
@brlogan does that last commit work for you?

brlogan:
I might not have time to actually test until Monday, but based on a quick review of the code, I think that will work!

mgoffin:
I did some more testing and I think it's good enough to go to master and iron out issues there. Thanks for the help!

brlogan:
Looks good, except I think we missed one thing.
If an Indicator already exists in CRITs with a set attack/threat type, and one adds that same Indicator to CRITs without choosing an attack/threat type, "Unknown" is added to the list.  Since adding an attack/threat type will remove "Unknown" from the list, it seems that we also shouldn't add "Unknown" to the list if the list already contains another option.

mgoffin:
I added that in this morning. There's a drawback to this, though. If an Indicator has a Threat/Attack type assigned that *isn't* `Unknown`, but realize they've made a mistake and want to set them back to `Unknown`, they cannot. The system has no way to separating "intentional" and "unintentional" addition of `Unknown` in these scenarios.

I suspect this might not be a big issue though. People would need to intentionally change the dropdown when adding an Indicator to something else, and I suspect if they messed that up it's because they picked the wrong non-Unknown value.

PR #793, opened 2016-08-03
Potential Improvements to Attack/Threat Type Lists

brlogan:
Rather than simply restricting removal of the last item in the list, when the user attempts to remove the last item, this will replace that last item with the `Unknown` option.
Also, if a user manually adds an item to the list in the Indicator Details view, this will automatically remove `Unknown` if it exists.  Previously this functionality only worked when adding from the Add Indicator form.
Further, this removes `Unknown` as an option in the suggestions list, but still allows it to exist as the only item in a list.
Used the `append` parameter to determine whether `Unknown` should be removed from an update to the list.

@mgoffin: Curious what you think about these changes. I think they are a little better for the user, but still fit with your vision.

### DIFF
--- a/crits/indicators/indicator.py
+++ b/crits/indicators/indicator.py
@@ -260,7 +260,7 @@ class Indicator(CritsBaseAttributes, CritsActionsDocument, CritsSourceDocument, 
         parsed_threat_types = [s.strip() for s in parsed_threat_types]
 
         unknown = IndicatorThreatTypes.UNKNOWN
-        if len(self.threat_types) and unknown in parsed_threat_types:
+        if len(self.threat_types) and unknown in parsed_threat_types and append:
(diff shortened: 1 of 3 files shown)


mgoffin:
Sounds pretty good! I'll check it out tomorrow!

mgoffin:
@brlogan Hmm, now that I've committed that I found another test and it seems to not be working.

Test:
- Indicator with Threat and Attack type of "Unknown"
- Attempt to add new Indicator which is identical to one above but with a non-Unknown Threat and Attack Type

Expected:
Both new Threat and Attack Types would be updated.

Actual:
Only Attack Type was updated.

brlogan:
Crap, that's my fault.  Somehow I left out a line.

PR #795, opened 2016-08-04
Vocabulary API Resource

brlogan:
This is an API resource that I created awhile ago to provide programmatic access to each CRITs vocabulary set.  We had a need to be able to pull the list of Indicator Types, in particular, so while I was at it, I made all of the vocabulary sets available.  It's not perfect, but gets the job done.  Figure I'd throw it out there so people could take a look at it.

The following URL path will return every vocabulary set:
`/api/v1/vocab/`

Providing a vocabulary class name, like below, will only return that particular set:
`/api/v1/vocab/IndicatorTypes/`

### DIFF
--- a/crits/core/api.py
+++ b/crits/core/api.py
@@ -751,3 +751,21 @@ def determine_format(request, serializer, default_format='application/json'):
 
     # No valid 'Accept' header/formats. Sane default.
     return default_format
+
+
+class MongoObject(object):
+    """Class that represents a Mongo-like object"""
+    def __init__(self, initial=None):
+        self.__dict__['_data'] = {}
+
+        if hasattr(initial, 'items'):
+            self.__dict__['_data'] = initial
+
+    def __getattr__(self, name):
+        return self._data.get(name, None)
+
+    def __setattr__(self, name, value):
+        self.__dict__['_data'][name] = value
+
+    def to_dict(self):
+        return self._data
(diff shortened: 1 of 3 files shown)


mgoffin:
Nice idea! Not sure what's causing it, but the default format should be JSON, not XML. This would make it consistent with the rest of the API.

brlogan:
Weird, I get JSON when I send a GET programmatically, but I get XML in the browser. I'll look into it.

brlogan:
@mgoffin : So it looks like this behavior is expected. JSON is the default, but the browser requests XML, so that's what the API provides: 
`http://django-tastypie.readthedocs.io/en/latest/debugging.html`
 Makes sense to me.  Do you still want me to change it?

mgoffin:
I'm confused. If I go to `/api/v1/indicators/` I get JSON. For every other API endpoint I get JSON as the default without issue. This is the only endpoint that is giving out XML by default. That seems wrong to me. I've never had to append `&format=json` to any requests to get JSON.

brlogan:
Found it. This snippet from /crits/core/api.py:
```python
def determine_format(request, serializer, default_format='application/json'):
    """
    This overrides the default tastyie determine_format.
    This is done because we want to default to "application/json"
    even though most browsers will send along "application/xml" in the
    Accept header if no "format" is provided.
```

I went ahead and changed the VocabResource to inherit from the custom CRITsAPIResource, which forces it to default to JSON.

mgoffin:
Sweet!

PR #799, opened 2016-08-06
bootstrap fixes

frbapolkosnik:
- Allow the bootstrap to run on Linux Mint
- use the fix for M2Crypto compile on more platforms
- Remove deprecated option: --allow-external


frbapolkosnik:
This is good to go.

PR #800, opened 2016-08-09
Fix secret generation

frbapolkosnik:
tr was not filtering properly, allowing things not in the specified set to go through


frbapolkosnik:
Tested on Linux and OSX, the output is on par with the original python script

PR #812, opened 2016-09-01
Switch out M2crypto with cryptography

frbapolkosnik:


apolkosnik:
cryptography is better supported than other crypto modules. Unlike both pycrypto and m2crypto, you can install cryptography on Windows without requiring C compiler.

PR #818, opened 2016-09-21
Django 1.10 support

apolkosnik:
This effort is aimed at getting CRITs to work on a recent version of Django, and allow backwards compatibility for the older mongoengine. It is synced with the latest master with ROLES.  

I've done a limited amount of manual testing, so if you wish to give it a try, I'd really appreciate it.

Since, there's great performance hit on anything above 0.8.x, I've baked-in backwards compatibility for old mongoengine. I've also forked the mongoengine and created a "0.8.9" which contains some small fixes that allow use of Django 1.9.x.

There are certain small suggestions:
- Try to stay on MongoDB 2.6.x or 3.2.x, maybe 3.4 will work with this, but the older version of PyMongo/Mongoengine might have some trouble keeping up, especially due to deprecation of some components.

- Mongoengine 0.8.x likes to be paired with PyMongo 2.8.1, this setup currently works with Django 1.8.x and 1.9.x

- Mongoengine 0.15 goes well with PyMongo 3.6, this setup works with Django 1.8.x - 1.11.x

This PR contains a whole slew of changes to make CRITs work on Django 1.8 - 1.11.
- updated all url.py files to use url() and import views, and referencing the view functions directly
  -Added name params to urlpatterns for compatibility with reverse() on 1.10 (RemovedInDjango110Warning: Reversing by dotted path is deprecated )
- Updated templates to reference view names rather than view class names
- RemovedInDjango110Warning: The context_instance argument of render_to_string is deprecated.
  Additionally, there are some other fixes and features included:
- fix the auth hack ( no more meta('pk') stuff
- security middleware and some additional security options added
- Added a listing of loaded top-level modules (only the ones without the "." in the name, and with **version** filled-in)
- fixed index creation, and removal via management command (previously some indexes were omitted)
- added -l option to test_install
- due to issues with Django 1.11, debug_toolbar has its own config option, keep in mind that you need to enable debug to see it too
- enabled -Wall in the scripts/server, so that the warnings are visible


apolkosnik:
Brought up to date with the latest master, and it works nicely with crits_services in my django_1.10_support branch

PR #825, opened 2016-10-03
Fix Adding related samples / attachments

robertsjw:
There was a problem with adding related samples to objects when the "Inherit Sources" checkbox was selected. Additionally, we should remove the "Add Attachment" to email button from Emails, and use the Add Related dropdown instead.

### DIFF
--- a/crits/emails/templates/email_detail.html
+++ b/crits/emails/templates/email_detail.html
@@ -67,7 +67,7 @@
                         <span data-type='{{ email_field.field_type }}' data-field='{{ email_field.field_name }}' title="Create Indicator: Warning: Indicator might already exist" style="float: right;" class="ui-icon ui-icon-circle-plus create-indicator"></span>
                     {% else %}
                         {% if not email_field.field_value or email_field.field_value.strip == "" %}
-                            <span data-type='{{ email_field.field_type }}' data-field='{{ email_field.field_name }}' title="Create Indicator: Warning: A value should be supplied first" style="float: right;" class="ui-icon ui-icon-alert create-indicator"></span>
+                            <span data-type='{{ email_field.field_type }}' data-field='{{ email_field.field_name }}' title="Create Indicator: Warning: A value should be supplied first" style="float: right;" class="ui-icon ui-icon-alert create-indicator"></span> 
                         {% elif email_field.field_value %}
                             <span data-type='{{ email_field.field_type }}' data-field='{{ email_field.field_name }}' title="Create Indicator" style="float: right;" class="ui-icon ui-icon-plusthick create-indicator">"{{email_field.field_value}}"</span>
                         {% endif %}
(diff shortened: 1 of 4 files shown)


apolkosnik:
No need for extra lines in 2 files.

crits/emails/templates/email_detail.html
```diff
@@ -68,7 +67,7 @@ <h3 class='titleheader'>
                         <span data-type='{{ email_field.field_type }}' data-field='{{ email_field.field_name }}' title="Create Indicator: Warning: Indicator might already exist" style="float: right;" class="ui-icon ui-icon-circle-plus create-indicator"></span>
                     {% else %}
                         {% if not email_field.field_value or email_field.field_value.strip == "" %}
-                            <span data-type='{{ email_field.field_type }}' data-field='{{ email_field.field_name }}' title="Create Indicator: Warning: A value should be supplied first" style="float: right;" class="ui-icon ui-icon-alert create-indicator"></span>
```
apolkosnik: White space change

crits/emails/templates/email_detail.html
```diff
@@ -122,7 +121,7 @@ <h3 class='titleheader'>
              {% include 'releasability_list_widget.html' %}
         {% endwith %}
         </tr>
-        </tbody>
```
apolkosnik: another white space change

crits/emails/templates/email_detail.html
```diff
@@ -235,4 +234,3 @@ <h3 class='titleheader'>
 {% block javascript_includes %}
     <script type="text/javascript" src="{{ STATIC_URL }}js/emails.js"></script>
 {% endblock %}
-
```
apolkosnik: one more white space change

brlogan:
I was hoping to comment on this one before it got merged, but I got distracted with other work.

Are we sure we want to get rid of the "Add Attachment" button?  Sure you can do the same thing with the "Add Related" menu, but the "Add Attachment" button simplifies the process and makes it consistent.  For example, it should ensure that the `Contains/Contained_Within` relationship is used, and it uses a specific Source Method to document what the analyst did.  Without that button it is less clear that the added Sample is indeed an attachment and not something else.

I don't think we want to force this behavior into the "Add Related" Sample handler because there may be cases where someone wants to relate to an email a Sample that is not an attachment.

mgoffin:
When you use "Add Related" you can simply change the relationship in the form to the one of your choosing. In this case, it would be "Contained Within" and then things work just like before, except now we don't have to manage two entirely separate interfaces to the same code (which was breaking things).

brlogan:
Yes, I realize this. It's just that, undoubtedly, analysts will occasionally forget to choose the correct relationship option, which will later lead to confusion.  And even if we ask them to use a specific Source Method, it won't be 100% consistent. They are all about reducing clicks and this adds some work, albeit a small amount.  The code could certainly use some cleanup to reduce the overhead of a separate interface, but I'm probably going to be asked to keep the button even if stays removed from public. : /

And actually, regardless of which relationship you pick in the "Add Related" Sample form, it always defaults to "Contained Within" if it's an Email, which isn't ideal. Of course, that old code would be easy to change.

Issue #811, opened 2016-08-31
AttributeError at /login/


obdnanr:
I upgraded from Ubuntu 14.04 to 16.04 and when I try to access the login page I get the following error:

`AttributeError at /login/

'NoneType' object has no attribute 'crits_message'

Request Method: 	GET
Request URL: 	http://server:8080/login/?next=/
Django Version: 	1.8.13
Exception Type: 	AttributeError
Exception Value: 	

'NoneType' object has no attribute 'crits_message'

Exception Location: 	/opt/crits-master/crits/core/views.py in login, line 366
Python Executable: 	/usr/bin/python
Python Version: 	2.7.12
Python Path: 	

['/opt/crits-master/crits',
 '/opt/crits-master',
 '/usr/local/lib/python2.7/dist-packages/pycrits-0.1-py2.7.egg',
 '/usr/lib/python2.7',
 '/usr/lib/python2.7/plat-x86_64-linux-gnu',
 '/usr/lib/python2.7/lib-tk',
 '/usr/lib/python2.7/lib-old',
 '/usr/lib/python2.7/lib-dynload',
 '/usr/local/lib/python2.7/dist-packages',
 '/usr/lib/python2.7/dist-packages']

Server time: 	Wed, 31 Aug 2016 11:17:45 -0400`

obdnanr:
Environment:


Request Method: GET
Request URL: http://chesva486test.dollartree.com:8080/login/?next=/

Django Version: 1.8.14
Python Version: 2.7.12
Installed Applications:
('crits.core',
 'crits.dashboards',
 'django.contrib.auth',
 'mongoengine.django.mongo_auth',
 'django.contrib.contenttypes',
 'django.contrib.messages',
 'django.contrib.sessions',
 'django.contrib.sites',
 'django.contrib.staticfiles',
 'crits.actors',
 'crits.campaigns',
 'crits.certificates',
 'crits.domains',
 'crits.emails',
 'crits.events',
 'crits.indicators',
 'crits.ips',
 'crits.locations',
 'crits.objects',
 'crits.pcaps',
 'crits.raw_data',
 'crits.relationships',
 'crits.samples',
 'crits.screenshots',
 'crits.services',
 'crits.signatures',
 'crits.stats',
 'crits.targets',
 'tastypie',
 'tastypie_mongoengine')
Installed Middleware:
('django.middleware.common.CommonMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
[...]

PR #827, opened 2016-10-04
Added a condition to check if CRITsConfig exists on Login Page #811

djsan15:
Basically the issue was that there were no CRITsConfig objects in the database.
So I have added a condition to check if there are no objects then it will create an objects with [ crits_message="Please Login to Proceed" ]. This message will be visible on login page.
**Here is screenshot of Login Page after this** 

screenshot from 2016-10-04 23 30 14

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,14 @@ Just run the following:
     sh script/bootstrap
 ```
 
-Once you've run bootstrap once, do not use it again to get the runserver going, you'll be going through the install process again. Instead use the server script:
+Once you've run bootstrap once, do not use it again to get the runserver going, you'll be going through the install process again.
+
+Run the management command to set the basic config ('*' to allow all the hosts):
+````
+	python manage.py setconfig allowed_hosts "*"
+````
+
+Use the server script to run the local server:
 
 ```bash
 
--- a/crits/core/views.py
+++ b/crits/core/views.py
@@ -342,6 +342,10 @@ def login(request):
 
     # Gather basic request information
     crits_config = CRITsConfig.objects().first()
+    #This will create a CRITsConfig Object if there are no objects in the DB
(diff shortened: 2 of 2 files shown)


mgoffin:
Howdy! One quick thing. If there's no CRITsConfig document in the database, it's very possible other stuff is broken and didn't get added to the database. That's all part of the installation process and shouldn't be skipped. Are you sure that wasn't skipped on your install?

djsan15:
Hey, I had followed the quick installation guide which says, run only 2 commands:
1. ```` sh script/bootstrap````
2. ````sh script/server````
These commands do not create a CRITsCongig.

But, the production document has the extra step required to create CRITsConfig:
````python manage.py setconfig allowed_hosts "foo" ````

I have pushed the changes in README.md and restored views.py to previous version.

mgoffin:
We have a step within bootstrap here which creates the default collections:

https://github.com/crits/crits/blob/master/script/bootstrap#L187

Within `create_default_collections` is a call to build the config in the database if it doesn't already exist (assumption is that if you already have a config in your database, you want to keep it and use setconfig to adjust it if necessary).

https://github.com/crits/crits/blob/master/crits/core/management/commands/create_default_collections.py#L54

That makes a call to a function within setconfig:

https://github.com/crits/crits/blob/master/crits/core/management/commands/setconfig.py#L146

So it seems the architecture for building the configuration in the database already exists and there's no need to add an extra step for the user. I suspect the issue for you was that the `bootstrap` did not complete properly and you didn't catch the error to see it. That error probably occurred before it would have created a config in your database.

I think the proper fix here is for boostrap to be more resilient to error, or at least have the ability to stop itself if there is an error, present it clearly to the user with directions on how to fix it, and be able to pick up where it left off when run again.

PR #831, opened 2016-10-10
Updates to bootstrap.

apolkosnik:
- Add support for Fedora
- Update the mongodb to 2.6.12
- try to update pip/setuptools to the latest


mgoffin:
Why isn't MongoDB 3.x being added instead of 2.6.x?

apolkosnik:
I didn't have the time to iron it out, new keys new URLs. I figured that the easiest might be to just point it to the latest 2.6.x

PR #837, opened 2016-10-25
Role Based Access Control

robertsjw:
This is a major change with how CRITs provides access to objects and will require a lot of testing to make sure I am not missing/breaking anything.  A few of the major changes:
- Roles
  - The Role object has boolean values that determine if a User has access to any object within CRITs. 
  - Sources are applied to Role objects so all users that are that role have access to data from Sources applied to that role. 
- Sources (and addition of TLP to SourceAccess)
  - SourceAccess objects are now assigned to Roles rather than individual users. 
  - Traffic Light Protocol (TLP) can be assigned to CRITs Objects as a part of a Source Instance

Added support for Role Based Access Control (RBAC). This is a major change with how CRITs provides access to objects. Instead of providing access solely based on a Source, it provides access based on the permissions defined in the Role object assigned to a user. Role objects then have sources that they have access to. When migrating from an existing CRITs Instance, you will need to take the following steps:
- Create default Roles and migrate legacy user roles to new objects:
    `python manage.py create_roles -a`
- Then, log in using an Administrator (now called UberAdmin) user to add appropriate SourceAccess to the new Role objects.
- Moving forward, SourceAccess objects need to be added to Role objects for users to have access to data from that source.


mgoffin:
Doing some testing. Lots of stuff looks good. Ran across a couple things:

1) With a user who shouldn't have access to the Control Panel, I was able to get to the role details page (I happened to have it in my `next` URL parameter at login time).
2) I get an error `Uncaught TypeError: Cannot read property 'length' of null(…)` when trying to click on the "Relationships Service" tab (my user has read-only access to the Actor TLO and that's it)
3) I got this error:
```
Exception Type:	UnboundLocalError
Exception Value:	
local variable 'role' referenced before assignment
Exception Location:	/Users/mgoffin/git/others/robertsjw/crits/crits/core/handlers.py in get_role_details, line 4387
```

I'll keep testing as there's a lot to test here!

mgoffin:
Also when trying to delete a relationship I get a `list index out of range` error.

mgoffin:
Sorry, testing has been limited for me lately. I want to try and test this "enough" so that we can merge with master and notify everyone what we are doing. The hope is that more people will help with testing and submitting issues/PRs that way. We'd do a final merge to stable_4 before doing this so anyone using master in prod can switch to stable_4 and not miss anything.

That being said, just got this error:

```python
File "/Users/mgoffin/git/others/robertsjw/crits/crits/core/role.py", line 120, in Role
    add_new_signature_dependency = BooleanField(dafault=False)
TypeError: __init__() got an unexpected keyword argument 'dafault'
```

mgoffin:
When logging in I am not redirected to my `next`, and if I don't have a `next` I am not redirected to the dashboard. End results is a successful login but staring back at the login screen.

mgoffin:
Would it be possible to do a merge with master to get up-to-date. I think we need to force this in and let things be fixed by everyone as they see stuff.

PR #851, opened 2016-11-03
Configure mention-bot with users to never notify

gtback:
Feel free to mention me directly if you'd like my input on anything 😀 

https://github.com/facebook/mention-bot#configuration

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -1,0 +1,3 @@
+{
+  "userBlacklist": ["gtback"]
+}


mgoffin:
Cool, so the bot looks for the config file in this repo, not the mention-bot that gets deployed to heroku?

gtback:
That's what I'm assuming, based on https://github.com/haskell/cabal/blob/master/.mention-bot and https://github.com/xbmc/xbmc/blob/master/.mention-bot. 

Also, [the code](https://github.com/facebook/mention-bot/blob/386c5ec0f03d9dc52c1473cb1325f2fd045e6833/server.js#L158-L165) seems to use the [GitHub API](https://developer.github.com/v3/repos/contents/#get-contents) to get the config from the repo.

PR #852, opened 2016-11-04
Wrapped template loaders in cached.Loader and added some nice debug toolbars

apolkosnik:
- Wrapped template loaders in cached.Loader  (https://docs.djangoproject.com/en/dev/ref/templates/api/#django.template.loaders.cached.Loader)
- Added some nice debug toolbars

### DIFF
--- a/UPDATING
+++ b/UPDATING
@@ -3,6 +3,16 @@ should be aware of. This will also include things like dependency changes or
 references to changes in other repositories. Updates are sorted by commit date
 descending.
 
+
+Author: apolkosnik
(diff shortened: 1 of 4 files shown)


apolkosnik:
debug toolbars only show up when DEBUG=True

apolkosnik:
Caching automatically turns off when DEBUG=True

mgoffin:
Let me know when you think you're all set with adding toolbar features :)

apolkosnik:
I think that it's enough for now

Issue #853, opened 2016-11-07
Relationships Collection 


armtash:
This is in continuation of #673 and the google groups discussion [here](https://groups.google.com/forum/#!topic/crits-development/u1Zsl3o3pS4)

Issues with current handling of relationships:

1. With the current method of forging relationships, a copy of the relationship is stored within the TLO documents.  This means that as you grow the number of relationships, you will also grow the size of the TLO documents.  I observed some very poor performance when attempting to forge more than a few hundred relationships between objects, due to the amount of time it was taking to load/save a document.  The fact that the relationships are stored within the TLO object is the greatest contributor to the performance issues.

2. Relationship Validation:  In order to validate that a relationship should be added or modified, a process takes place in "crits_mongoengine.add_relationship" or "crits_mongoengine._modify_relationship",  which involves loading all existing relationships, and then iterating through them to validate/modify the relationship.  With a few hundred relationships, this is a reasonable operation, however when you start attempting to add a relationship to a TLO with 10,000 existing relationships, this method of validation is not ideal.
With a relationship collection, you can quickly validate or find a relationship by performing a mongo query which asks, "Is there relationship document that exists which has LEFT_OBJ and RIGHT_OBJ with REL_TYPE X?"

3. Relationship Sorting: "crits_mongoengine.sort_relationships" is most leveraged in order to sort the relationships by TLO before rendering.  Again, the issue is related to the fact that we are attempting to load all relationships, and then sorting them by TLO so that we can pass the relationships to be rendered by the details template.  With a relationship collection, rather than loading everything and then sorting, you can query for the relationships with X TLO type.

For Example:
sorted_relationships = {}
sorted_relationships['Indicators'] = Relationships.objects("Find all relationship documents with LEFT_OBJ SELF, and RIGHT_TYPE Indicators")
sorted_relationships['Domains'] = Relationships.objects("Find all relationship documents with LEFT_OBJ SELF, and RIGHT_TYPE Domains")

4. Rendering: At the moment, even if we improve the way that relationships are stored/added/modified/validated, there is still an issue with rendering thousands of relationships.  The way that the user views/interacts with the relationships will need to be improved in the future.

Potential Changes:

1. A new relationship_collection is created, and all relationships are migrated from the TLOs to this collection.  This will require a schema change to every CRITs TLO.

A relationship document might look like so:

{
    _id: ObjectId(""),
    left_obj: { 'obj_id': 'ID of left object',
               'obj_type': 'Type of left object',
[...]

Issue #673, opened 2015-12-17
Performance issue with TLOs and many relationships


armtash:
I have tested this out with the latest master commit on two servers.

As the number of relationships for a TLO grow, the amount of time it takes to add an additional relationship grows significantly.

I currently have a raw_data object, and I have forged around ~350 (out of 800) relationships with different Indicators.  I am forging these relationships using the patch API calls.

When attempting to forge a relationship using the web interface, my last request took 40 seconds to complete.  The API calls are taking just as long.

mgoffin:
Possibly all of the looping it's doing to see if the relationship already exists to prevent duplicates.

dsnellgrove:
We have some items with tens of thousands of relationships. To get all these on there we had to modify the code to allow bulk relationship creation. When I profiled the original code I found that the loops were actually very fast and instead the majority of the time was spent in the mongoengine module converting the mongo object to python and back again with each call; I had expected the loop to be the slow code but was wrong. We added some code to short-circuit this process during bulk creation and instead reuse the object so there was one read and one write with many relationships created in the middle.

This was quite some time ago, but before trimming the loop profile the code, I wasted a lot of time trying to optimize the wrong code at first to be disappointed the issue was in the underlying module :(

armtash:
dsnellgrove, could you provide any more details? Can you find exactly what you changed?

webstergd:
We I think made a similar change for the sample api calls. I think this [1]
might be what you are looking for. We stopped pushing any changes to the
API until it finished moving to the new format.

[1] https://github.com/crits/crits/pull/620

thelok:
@dsnellgrove How did you get around the 16MB limit for documents if you have that many relationships? Unless you changed the architecture of how relationships are stored?

dsnellgrove:
They're still squeezing into the 16MB limit. I think there was one I found that had hit it though--further additions produced errors since the indicator couldn't be saved after the edit.

Issue #854, opened 2016-11-08
Replace MongoEngine


mgoffin:
Since starting use of it within the project, the featureset and performance have gone downhill. It's better we replace this with something of our own. More detailed information about why things have gone downhill can be found here:

https://github.com/marrow/contentment/issues/12

My suggestion for replacement is to redo the backend code to support interchangeable databases. Configure which database you want to use and the classes should handle doing the right thing. This will allow those folks on MongoDB to stay, and those folks who want to be on ElasticSearch to do so.

This would impact the API as well which should be cleaned up anyways given how slow it is with the multitude of layers you have to pass through to do stuff.

mgoffin:
Thinking through this I had some ideas. The first one being the class everything inherits to gain its ability to perform database operations. I was thinking we would setup `settings.py` to allow for people to choose between the databases we have support for, so something like:

```python
DATABASE_TYPE='MongoDB'
#DATABASE_TYPE='ElasticSearch'
```

We would then have a `backend` folder which would contain a file for each backend database we wish to support. A file `crits_mongodb.py` might contain:

```python
class CRITsMongo(object):
    # c will be set to the pymongo cursor (or maybe a list?) that is returned for iteration 
    c = None
    # l will be the length of the cursor returned
    l = 0
    # The collection to query which will be overwritten by whatever class inherits this one
    coll = None

    def __init__(self, c=None):
        if c:
            self.c = c
        ....
    def objects(self):
        """ Query for content in the database""""
        ....
    def save(self):
        """ Iterate over "c" and save the content back to the database """
        ....
    def delete(self):
        """ Iterate over "c" and delete the content in the database """
        ....
```

And we would have the main class be something like:

```python
from django.conf import settings

from backend.crits_mongodb import CRITsMongo
from backend.crits_elasticsearch import CRITsElasticSearch

if settings.DATABASE_TYPE == 'MongoDB':
    backend_db = CRITsMongo
elif settings.DATABASE_TYPE == 'ElasticSearch':
    backend_db = CRITsElasticSearch

class CRITsDB(backend_db):
     def __init__(self):
        ....
```

This would allow us to create a modular backend for supporting more than one database, but still provide similar functionality for working with the data once it is in python-land. My theory for setting `c` for the `CRITsMongo` class is so you can do something like this (let's assume `Sample` inherits `CRITsDB`):

```python
from crits.samples.sample import Sample

samples = Sample.objects(query={'campaigns.name': 'foo'}, filters={'limit': 1})
[...]

PR #864, opened 2016-11-29
Relationship collection

armtash:
This PR is in reference to #853, #673  and the google groups discussion [here](https://groups.google.com/forum/#!topic/crits-development/u1Zsl3o3pS4).

In short, the changes associated with this PR will alter the way that relationships are stored within CRITs. A new relationship collection will be leveraged for storing relationships, as opposed to storing relationships within a TLO document.  The implementation is similar to how comments are currently stored in CRITs.  The changes should allow you to forge significantly more relationships (100k+).

Please keep in mind that while these changes might allow for the forging of a significant number of relationships, additional changes will need to be made in order to render these properly to the end user.  Changes to how relationships should be rendered has been discussed in #853, and also [here](https://groups.google.com/forum/#!topic/crits-development/u1Zsl3o3pS4).

Another thing to note is that pymongo was intentionally leveraged as much as possible, to avoid any potential perfomance issues with MongoEngine. 

The schema for a relationship document looks like so:
```
{
    _id: ObjectId(""),
    left_obj: { 'obj_id': 'ID of left object',
               'obj_type': 'Type of left object',
               'rel_type': 'Type of left relationship',
               },
    right_obj: { 'obj_id': 'ID of right object',
               'obj_type': 'Type of right object',
               'rel_type': 'Type of right relationship',
               },
    date: 'ISODate when this relationship was created',
    relationship_date : 'ISOdate of when this relationship occurred.',
    analyst: 'The analyst, if any, that made this relationship',
    rel_reason: 'Reason for the relationship',
    rel_confidence: 'Confidence of the relationship',
}
```
**Creating Relationships:**
The creation of relationships should be no different to what it was like before.  You can still call tlo.add_relationship(), which will then leverage relationships.handlers.forge_relationship().
**Modifying Relationships:**
You can no longer call tlo._modify_relationship() or functions like tlo.edit_relationship_reason().  These functions have been moved to relationships.handlers.edit_relationship_reason() and relationships.relationship.modify_relationship().
**Pulling Relationships:**
You can no longer leverage the attribute tlo.relationships, as it is no longer a valid attribute for a TLO.  In order to pull the relationships associated with a TLO, you can call tlo.get_relationships(), which then calls relationships.handlers.get_relationships().  
**Sorting Relationships:**
tlo.sort_relationships() has been merged into get_relationships().  You can define how you want relationships returned like so: get_relationships(sorted=True, meta=True)
**Sanitizing Relationships:**
The logic behind tlo.sanitize_relationships() has also been merged into get_relationships.  If a username is passed to get_relationships, the results will be sanitized based on what they have access to.  If no username is passed, the results will include everything.

When interacting with relationships (ie. indicator_obj.get_relationships()), it is important to note that the dictionary that represents a relationship will resemble something like this:

```
{
    _id: ObjectId(""),
    other_obj: { 'obj_id': 'ID of right object',
               'obj_type': 'Type of right object',
               'rel_type': 'Type of right relationship',
               },
    date: 'ISODate when this relationship was created',
    relationship_date : 'ISOdate of when this relationship occurred.',
    analyst: 'The analyst, if any, that made this relationship',
    rel_reason: 'Reason for the relationship',
    rel_confidence: 'Confidence of the relationship',
}
```
Note that "other_obj" will represent the "right_obj", and that "left_obj" is excluded.

**Migration:**
If the migration to the relationship collection fails, the failed relationships will be saved in the "unsupported_attributes" attribute.  However, if the relationship creation fails due to one of the following conditions, the error is ignored:
1. Right TLO can't be found (Represents a broken relationships pointing to a TLO that no longer exists).
2. Relationship already exists (Relationship which has already been created by the other related TLO).

Note that in a scenario where relationships are no longer in sync (relationship_date is not identical), two relationships will be created.  This is mentioned in greater detail in #853.


ckane:
I was just looking at some other issues raised by the two-part relationships in CRITs and was thinking the same thing. Glad to see someone else is ahead of me on the work on this effort. I'll try to take some time in the near future to look at this, play with it and investigate some of the conflicts against master that exist now.

armtash:
Around the same time I had submitted this PR, there was another change proposed to entirely removing mongo and making the DB layer more interchangeable, issue #854, so that you could potentially implement some RDBMS rather than mongo if desired.  I'm not sure if there has been any progress with that, but I think that is the right approach.  My changes are helpful only in the short term.

Issue #872, opened 2016-12-09
Docker file


imaia:
What about a docker file for easy start?

PR #879, opened 2017-01-19
Create Dockerfile

moshekaplan:
Created a Dockerfile with crits and its dependencies.

All the required components (crits, MongoDB, ...) are included in a single Docker container. No additional configuration is needed. The docker-based crits instance can be started with the following command:

`docker run --publish=8080:8080 crits/crits:latest`

The instance can then be accessed on `localhost:8080`

The default username is: `admin`
The default password is `pass1PASS123!`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM ubuntu:latest
+
+MAINTAINER crits
+
+RUN apt-get -qq update
+# git command
+RUN apt-get install -y git
+# pip command
+RUN apt-get install -y python-pip
+# lsb_release command
+RUN apt-get install -y lsb-release 
+# sudo command
+RUN apt-get install -y sudo
+# add-apt-repository command
+RUN apt-get install -y software-properties-common
+
+RUN git clone --depth 1 https://github.com/crits/crits.git 
+
+WORKDIR crits
+RUN TERM=xterm sh script/bootstrap < docker_inputs
+
+EXPOSE 8080
+
+CMD sh contrib/mongo/mongod_start.sh && python manage.py runserver 0.0.0.0:8080
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN apt-get install -y sudo
 RUN apt-get install -y software-properties-common
 
 # Clone the repo
-RUN git clone --depth 1 https://github.com/moshekaplan/crits.git 
+RUN git clone --depth 1 https://github.com/crits/crits.git 
 
 WORKDIR crits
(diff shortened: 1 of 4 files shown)


mgoffin:
Thanks for contributing! Could you modify the description of this PR to go over what this does, how it works, etc.? For people not using Docker and wanting to try it out, it would help having a baseline to understand what's going on when testing to ensure it's working like expected.

Thanks again!

moshekaplan:
Related issue: #872 .

I only noticed one issue with this dockerfile: The password is randomly generated when the bootstrap script runs. That means a randomly-generated hardcoded password will be shared for all instances of the built docker container. This is bad because:

1. Hardcoded passwords aren't secure (not a huge issue for a Docker container only used locally for testing)
1. Randomly-generated passwords are not user friendly.

moshekaplan:
@mgoffin : Sorry, the updated message should be better.

moshekaplan:
I think that a hardcoded admin password is acceptable for a Docker image on non-production use.

Would it be possible to modify crits to allow specifying passwords? According to what I'm seeing here, [core/management/commands/users.py#L123](https://github.com/crits/crits/blob/master/crits/core/management/commands/users.py#L123) , it seems that passwords cannot be set directly.

mgoffin:
If someone wants to make the changes to allow for providing a password as well as being assigned, that's fine.

moshekaplan:
Modified to use the hardcoded password of `pass1PASS123!`

mgoffin:
Thanks! If there are issues hopefully people will submit them here or in Slack and we can maintain accordingly.

wzod:
Nice work, @moshekaplan!

Regarding concern with password, maybe a possible solution may be similar to the one used for the REMnux dockerfile for CRITs:

https://github.com/REMnux/docker/tree/master/CRITs

The idea was to still use the password generation routine and present that to the user when starting the instance. Not sure if that resolves the concern, but figured to share in case that helps.

moshekaplan:
Thankws @wzod. If I had known there was an existing Dockerfile, I probably wouldn't have written this one.

I see that the REMNUX Dockerfile resets the password in the final ENTRYPOINT command. It doesn't look like a bad approach either. I'd need to dig into seeing exactly how that works in practice - like would it reset the password each time the image was restarted?

Either way, I'm not certain what is the better approach in this situation: A hardcoded pass or generating a new one for each run. I leaned towards the usability of the first. The best would be to randomly generate a password the first time a user launches the image, but never again afterwards. I'm not familiar enough with Docker to know how that could be implemented.

PR #893, opened 2017-04-03
Fixed API problem for indicator activity

ericwentz:
Fixes two problems regarding the indicator activity API functions. The activity_add and activity_update functions in this handlers.py module worked fine from the web UI, but failed from the authenticated API due to the different passed parameters.  This change fixes that problem by recasting the "activity" variable which was passed into the function when called from the api


mgoffin:
Concerned about why the entire `indicators/api.py` file is showing as changed. I don't think it should be like that.

mgoffin:
Also, it seems that if you are performing at PATCH request, all you need to do is pass `activity` as a dictionary and there would be no changes required here?

ericwentz:
Not sure why the entire api.py shows up as changed - that one was the minor one that we talked about last week - not nearly as important as the one from the PATCH api call - the only line changed in the api.py file was line 161 where the 3rd argument was added to the call.

ericwentz:
I cannot pass activity as a dictionary as this comes from the crits.core.api.py file - if i change the calling format, I may break any other calls out of the core.api.py file - I had to fix it on the receiving end.

ericwentz:
Just curious - what should I expect regarding this PR?

mgoffin:
I used the example Python code on the Wiki for performing a PATCH request, and modified it to perform an `activity_add` action and it worked perfectly without having to include any of the changes in this PR:

```python
#!/usr/bin/env python

  import datetime
  import json
  import requests

submit_url = 'http://localhost:1337/api/v1/indicators/583c94fbd1f467c74df772b3/'

  params = {
    'api_key' : 'REDACTED',
    'username' : 'mgoffin',
  }

  headers = {
    'Content-Type' : 'application/json',
  }

  # date must be in the format %Y-%m-%d %H:%M:%S.%f
  formatted_date = datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S.%f')

  data = {
      'action': 'activity_add',
      'activity': {
          'description': 'Test activity add!',
          'date': formatted_date,
          'start_date': formatted_date,
          'end_date': formatted_date,
      }
  }

  r = requests.patch(submit_url, headers=headers, params=params,
                data=json.dumps(data), verify=False)

  if r.status_code == 200:
      print('Yay!')
```

mgoffin@smoke > ~/git/crits/crits > master > python ~/python/crits_api_test.py
Yay!
mgoffin@smoke > ~/git/crits/crits > master >

I looked at my local instance and the activity was added as expected.

ericwentz:
Thanks Mike!  The docs are not 100% clear on how to form this patch request.  I hope others are able to use your example above to eliminate any future headaches!

mgoffin:
This was a literal copy/paste from the documentation, with changes for the
URI to hit and the payload of the request.

PR #894, opened 2017-04-07
This change makes CRITs work on Mongoengine 0.12

apolkosnik:
This thing broke after https://github.com/MongoEngine/mongoengine/pull/1497 was merged.
I'm not really sure if this is the right fix, it seems that mongoengine/fields.py#L625 added a check for 'dict'.

### DIFF
--- a/crits/services/analysis_result.py
+++ b/crits/services/analysis_result.py
@@ -88,7 +88,7 @@ class AnalysisResult(CritsDocument, CritsSchemaDocument, CritsDocumentFormatter,
     log = ListField(EmbeddedDocumentField(EmbeddedAnalysisResultLog))
     object_type = StringField(required=True)
     object_id = StringField(required=True)
-    results = ListField(DynamicField(dict))
+    results = ListField(DynamicField())
     service_name = StringField()
     source = StringField()
     start_date = StringField()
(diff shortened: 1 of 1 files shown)


apolkosnik:
I think it works, I've tested it by kicking off couple of services, and I've not noticed any issues/differenece.

apolkosnik:
Ok, so we were kind of lucky that it worked at all... See
https://github.com/MongoEngine/mongoengine/issues/1542#issuecomment-299747060

mgoffin:
Ok. So should we leave "dict" there or remove it entirely?

apolkosnik:
My understanding is that DynamicField() will be equivalent to DynamicField(dict())

apolkosnik:
Just less typing

mgoffin:
But only in the new versions?

apolkosnik:
I've tested with results = ListField(DynamicField()) on mongoengine versions 0.8.8, 0.10.6, and 0.13.0. It seems to be working ok. One more thing that needs to be addressed is a changes that need to be done to taxii_service.

apolkosnik:
https://github.com/crits/crits_services/pull/315 goes with this. That pull also fixes import for more recent versions of taxii (I'm not addressing addition of mixbox and whatnot)

PR #897, opened 2017-04-20
Provides configuration for Apache 2.4 (Ubuntu/Debian)

execthis:
Provides working configuration running CRITs on Apache 2.4 for Ubuntu and Debian. We had a lot of trouble with the existing provided configuration which is i believe for apache 2.1. 

Regards 
Remo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ INSTALL_RECEIPT.json
 
 # Ignore VirtualEnv
 venv/*
+
+# jetbrains ide
+.idea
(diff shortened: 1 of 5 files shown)


execthis:
Btw. where can i find the wiki files for "Production grade CRITs install" to update them as well? Thanks

mgoffin:
Is there a specific reason why you want to move what we consider the "default" directory of `/data/crits` to something else? Functionally it shouldn't matter that things are in `/data/crits` as long as configuration is properly set.

execthis:
its just a symlink who points to /data/crits/extras/www. everything is still located under /data/crits. Otherwise I had to create a directory section for every alias or dir inside the virtual host because of "Require all denied" on root from apache2.conf:
```
<Directory />
 	Options FollowSymLinks
 	AllowOverride None
 	Require all denied
 </Directory>
```

I tried to give access to them like this with no success:
```
<Directory "/var/www/crits/(static|js|css|ext|images|img|new_images|plugin)">
       Options Indexes FollowSymLinks MultiViews
       AllowOverride All
       Order allow,deny
       allow from all
</Directory>
```

Im wondering why Directory directives are not recursively valid? So I decided to create a symlink to prevent having a lot of directory directives inside the virtual host configuration. Do you have a better approach?

Thanks

mgoffin:
I don't. I'm just trying to play devil's advocate here to determine if the symlink having to be created by the admin of the box is a necessary step as opposed to us having a configuration which already points to the necessary directories where the admin is expected to explode the code base.

execthis:
I'll try again tomorrow without the symlink. I just read that directory directives are applied recursively... normally: http://httpd.apache.org/docs/current/sections.html#file-and-web

PR #899, opened 2017-04-24
Adding feature to check to see if the sample is a Office file.

echin20:
Adding a feature to check if a sample is a Office file or not based on the file header hex bytes. Hopefully this is the correct way to check against the sample in CRITs.


mgoffin:
For clarification, what versions of Office files does this detect?

echin20:
This should only detect on the older format of Office (.doc, xls, ppt, etc.). The newer Office format are actually zip files.

PR #906, opened 2017-05-24
Record User for TLO Creation/Modification & Status Modification

brlogan:
Sometimes it is not always clear which user created a particular TLO and who last modified a TLO.
Additionally, from a workflow perspective, it is important to know who last modified a TLO's status field.
This change sets a few new values in the database to record this information for all TLOs.  Because these are new fields, no migration is necessary.

While implementing this, I realized that the display of creation/modification timestamps on TLO detail pages is very inconsistent. Some show both, while others show just one or neither.  This PR does not change which TLOs display those values, but I think it might be valuable to discuss making this consistent across all TLOs.

I'm open to feedback. Let me know if you notice that I missed something!

### DIFF
--- a/crits/actors/templates/actor_detail.html
+++ b/crits/actors/templates/actor_detail.html
@@ -64,7 +64,7 @@
                     </tr>
                     <tr>
                         <td class="key">Created</td>
-                        <td>{{actor.created}}</td>
+                        <td>{{actor.created}}{% if actor.created_by %} by {{ actor.created_by }}{% endif %}</td>
                     </tr>
                     <tr>
                         <td class="key">Modified</td>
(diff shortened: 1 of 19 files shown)


mgoffin:
I'm fine adding exposure of both created and last modified dates and times to this PR. It seems essential to see both if you are trying to reference who and when!

brlogan:
Okay cool.  The last items in the table for most TLOs are (in order): Status, Sectors, Sources (where applicable), Releasability.  What do you think about making every TLO's summary table end with:

Created
Modified
Status
Sectors
Sources (where applicable)
Releasability

mgoffin:
Seems consistent!

brlogan:
Okay! All TLOs should now display Created and Modified timestamps.  I changed the order slightly (moved Sectors before Created) because it looked better in a few cases.  The code and presentation is more consistent now. :)

Issue #660, opened 2015-11-23
Error uploading file on Microsoft Outlook .MSG File


myntric:
We have had a significant increase in phish messages arriving that contain embedded code that CRITs seems to fail processing each time.  Below is a snippet of the error code logged with Apache2.  This has been an ongoing issue for awhile.  

Mon Nov 23 10:12:37.743647 2015] [:error] [pid 8527:tid 140503714576128] 3092311 ERROR: Internal Server Error: /emails/new/outlook/
[Mon Nov 23 10:12:37.743816 2015] [:error] [pid 8527:tid 140503714576128] Traceback (most recent call last):
[Mon Nov 23 10:12:37.743888 2015] [:error] [pid 8527:tid 140503714576128]   File "/usr/lib/python2.7/dist-packages/django/core/handlers/base.py", line 112, in get_response
[Mon Nov 23 10:12:37.743898 2015] [:error] [pid 8527:tid 140503714576128]     response = wrapped_callback(request, *callback_args, **callback_kwargs)
[Mon Nov 23 10:12:37.743905 2015] [:error] [pid 8527:tid 140503714576128]   File "/usr/lib/python2.7/dist-packages/django/contrib/auth/decorators.py", line 22, in _wrapped_view
[Mon Nov 23 10:12:37.743912 2015] [:error] [pid 8527:tid 140503714576128]     return view_func(request, *args, **kwargs)
[Mon Nov 23 10:12:37.743918 2015] [:error] [pid 8527:tid 140503714576128]   File "/data/crits/crits/emails/views.py", line 444, in email_outlook_add
[Mon Nov 23 10:12:37.743925 2015] [:error] [pid 8527:tid 140503714576128]     ticket)
[Mon Nov 23 10:12:37.743951 2015] [:error] [pid 8527:tid 140503714576128]   File "/data/crits/crits/emails/handlers.py", line 783, in handle_msg
[Mon Nov 23 10:12:37.743959 2015] [:error] [pid 8527:tid 140503714576128]     result = parse_ole_file(data)
[Mon Nov 23 10:12:37.743966 2015] [:error] [pid 8527:tid 140503714576128]   File "/data/crits/crits/emails/handlers.py", line 1518, in parse_ole_file
[Mon Nov 23 10:12:37.743973 2015] [:error] [pid 8527:tid 140503714576128]     email['raw_body'] = get_stream_data(entry).decode('utf-16')
[...]

PR #916, opened 2017-06-20
Added encoding detection for email attachment uploads

auxiliumsec:
Fixes issue found in #660 

Automagically detects the encoding of an email and decodes accordingly. Unfortunately adds a new dependancy. Supports multiple different character encoding sets

Potentially useful in non .msg email uploads


apolkosnik:
If you could fix a conflict in requirements, it would be great!

auxiliumsec:
Resolved conflict as requested

PR #929, opened 2017-09-02
Fix and improve API queries for c-_id on TLOs

ckane:
One of our team realized that c-_id queries in the API were broken (a small if/elif swap bug). When digging deeper, we decided it would be nice to implement the __in and __nin API operator extensions here, as the tastypie /set/ endpoint is painfully inefficient for fetching a specific list of indicators by id.

So - fixed the ?c-_id bug, and then added support for c-_id__in and c-_id__nin to the API.

### DIFF
--- a/crits/core/api.py
+++ b/crits/core/api.py
@@ -449,7 +449,7 @@ class CRITsAPIResource(MongoEngineResource):
                     querydict['_id'] = ObjectId(v)
                 except:
                     pass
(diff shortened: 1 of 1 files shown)


ckane:
I've had this PR open for a bit and running on my dev instance for awhile. Wanted to know if you mind if I pull it in. Seems to be working for me & team so far.

PR #930, opened 2017-09-02
Optimization: Run bulk queries, per rel. type

ckane:
Rather than performing a mongoengine lookup on each TLO id in a
relationship list (the 'value' field in the BSON), group the TLO Ids
together in a bulk lookup using id__in, and run exactly one search
against each collection to fetch the population of related indicators
in that collection. This hopefully will speed up page rendering where
a TLO has many (> 1000's) of relationships.


ckane:
I've had this PR open for a bit and running on my dev instance for awhile. Wanted to know if you mind if I pull it in. Seems to be working for me so far.

PR #937, opened 2017-09-21
Fix bug on indicator

lprat:
- Fixed bug on indictor add by dataminer
- Fixed bug on indicator update (problem with object user and string user.username)
- Fixed bug to add object in indicator (need tlp information to add and problem with user.username)

Thanks!
Lionel

### DIFF
--- a/crits/indicators/handlers.py
+++ b/crits/indicators/handlers.py
@@ -1080,7 +1080,7 @@ def activity_add(id_, activity, user, **kwargs):
                 'message': 'Could not find Indicator'}
     try:
 
-        activity['analyst'] = user
+        activity['analyst'] = user.username
         indicator.add_activity(activity['analyst'],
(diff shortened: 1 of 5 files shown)


lprat:
add: Fix bug in parse ole date email and user.username relationship

mgoffin:
Debating whether we want to still allowing uploading if creation of the ISODATE fails. Thoughts?

lprat:
You have right. In my first patch, i was not understand than the problem came of olefile. The real problem is 0x00 add by olefile on header email, byte null prevent to parse header.

mgoffin:
You are adding a lot of stuff to this. I think it would be best if this PR focused on the original specific issue. Adding new Indicator vocabulary is usually an issue submitted with a request to add it and what it would be used for to ensure it's unique and not already provided.

PR #968, opened 2018-03-08
Use pymongo query to speed things up a bit

apolkosnik:


brlogan:
I've had to do similar things in the past.  Using straight Pymongo instead of Mongoengine can be massively faster.

apolkosnik:
In this case due to TLP we are running through the whole collection and evaluating TLPs for all of the TLPs... I think that there are a few sneaky ways to avoid this. This was sort of a PoC. The next step is try to try this in some other places and then use
```DocClass._from_son(raw_pymongo_dict)``` to create a single instance of a document from raw pymongo dict

PR #974, opened 2018-03-14
Save a db query for user_context

apolkosnik:
Since we already have user in request, there's no need to query the DB.

### DIFF
--- a/crits/core/views.py
+++ b/crits/core/views.py
@@ -1390,9 +1390,19 @@ def user_context(request):
     """
 
     context = {}
-    # Get user theme
-    user = CRITsUser.objects(username=request.user.username).first()
-    if user:
+    user = request.user
+    if not user.username:
+        # Get user theme
+        user = CRITsUser.objects(username=request.user.username).first()
+        if user:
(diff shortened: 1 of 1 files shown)


crits/core/views.py
```diff
@@ -1390,9 +1390,19 @@ def user_context(request):
     """
 
     context = {}
-    # Get user theme
-    user = CRITsUser.objects(username=request.user.username).first()
-    if user:
+    user = request.user
+    if not user.username:
+        # Get user theme
+        user = CRITsUser.objects(username=request.user.username).first()
```
mgoffin: How can we use `request.user.username` when we just checked if we don't have `user.username` which is the same thing?

crits/core/views.py
```diff
@@ -1390,9 +1390,19 @@ def user_context(request):
...
+        if user:
```
mgoffin: Why is this entire block of code duplicated? There should be a better way to do this without duplicating the code or building another function. Seems like the condition checks here are off and should be done differently.

PR #976, opened 2018-03-20
Speed up data query by using pymongo pass-through

apolkosnik:


apolkosnik:
I've got to revert couple of commits in this three, and resubmit.

PR #977, opened 2018-03-20
Fast queries via pymongo pass-through

apolkosnik:
This PR changes data_query() to use pymongo, and removes some needless operations in dashboard.
It appears to be working on the older and the newer versions of pymongo.

There's still considerable speed difference between different versions of mongoengine/pymongo
According to the timer at the bottom, I got the following results:

mongoengine 0.8.9 / pymongo 2.8.1
dashboard 3.6s
sample list 2.196s
sample detail 2.96s

mongoengine 0.15/ pymongo 3.6.1
dashboard 12.698s
sample list 2.95s
sample detail 5.092


apolkosnik:
got to close this one

PR #979, opened 2018-03-26
Try to avoid issues with pymongo and connection pooling 

apolkosnik:
This PR extends #978, and tries to avoid issues with pymongo and connection pooling when SERVICE_MODEL is using multiprocessing. More details are at: http://api.mongodb.com/python/current/faq.html#using-pymongo-with-multiprocessing.

Internal queries in CRITs now go through the new internal functions _mongo_connector() and _gridfs_connector() in mongo_tools.py, and the other few functions are implemented, so that they take advantage of the persistent connection to the database.

The old functions are modified so that depending on the SERVICE_MODEL they either kick off the separate connections (for process or process_pool) or otherwise the persistent connection is used.

These changes should make the things transparent for the services, and route the core's queries through the persistent connection.


apolkosnik:
It looks like pymongo 3.x doesn't like my trickery.

PR #984, opened 2018-04-24
Handle non-ASCII in changed field notifications

brlogan:
Modifying a TLO leads to the generation of an audit notification listing which fields were changed, and in the case of a list field, which values were added/removed.  These values are forced to Unicode, but no encoding is given. If the added/removed values contain non-ASCII characters, the Unicode Decode fails. This fix tells it to use UTF-8.

### DIFF
--- a/crits/notifications/processor.py
+++ b/crits/notifications/processor.py
@@ -109,9 +109,15 @@ class ChangeParser():
 
         message = ""
         if len(added_names) > 0:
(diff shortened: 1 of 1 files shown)


brlogan:
Please don't accept this yet. I want to update with a couple other cases.

brlogan:
Okay, I think that's good enough for this one.

PR #991, opened 2018-05-11
Add some optional caching

apolkosnik:
Also, commented out dummy cache, and 
updated some comments, and added a reference article for caching. It'll make it easier to hook CRITs up to memcached.

### DIFF
--- a/crits/config/config.py
+++ b/crits/config/config.py
@@ -41,6 +41,7 @@ class CRITsConfig(CritsDocument, Document):
     csv_max = IntField(default=25000)
     debug = BooleanField(default=True)
     enable_dt = BooleanField(default=False)
+    enable_dj_cache = BooleanField(default=False)
     depth_max = IntField(default=10)
     email_host = StringField(default='')
     email_port = StringField(default='')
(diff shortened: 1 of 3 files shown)


mgoffin:
What sort of stuff does this cache? Curious if we somehow leak anything we shouldn't because it got cached by a different user.

apolkosnik:
AFAIK this does a whole site caching. You are right, it's probably better to get a per user caching.
I was looking at https://stackoverflow.com/questions/20146741/django-per-user-view-caching 
One of the options there are:
https://gist.github.com/caot/6480c39453f5d2fa86bf
and
https://djangosnippets.org/snippets/2524/

mgoffin:
Ahh so we would have to define decorators and apply them to every single view. That seems cumbersome, especially if we want this to be optional and something to `enable`.

apolkosnik:
Hopefully it's more palatable now.

apolkosnik:
- adjustments to the middleware ordering for caches, and conditionals for middleware loading
- updated some comments
- uncomented the local mem cache backend, as it is in master 
- added option for turning on/off django caching (whether to load or not the middleware)

apolkosnik:
@mgoffin Please let me know if that's better

PR #997, opened 2018-05-18
Include "user" parameter in "activity_add" function call

carterchurch:
The API call to add data to the 'activity' field in Indicators (by POSTing to api/v1/indicator_activity) is currently not working properly because it is missing the required parameter "user"; Simple commit to add that parameter to the activity_add function call (see /crits/crits/indicators/handlers.py line 1061)

### DIFF
--- a/crits/indicators/api.py
+++ b/crits/indicators/api.py
@@ -170,7 +170,7 @@ class IndicatorActivityResource(CRITsAPIResource):
                     'end_date': end_date,
                     'description': description,
                     'date': datetime.datetime.now()}
-        result = activity_add(object_id, activity)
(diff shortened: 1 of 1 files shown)


mgoffin:
I’m not at a machine to check, but I believe this feature needs to be removed from the Indicator POST and moved over to the PATCH (if it isn’t already there).

mgoffin:
No problem! Check the core/api.py file for actions and see if this feature is listed there. If it isn’t, we just need to expose the handler and do a little cleanup on it to be compliant with the API PATCH usage.

carterchurch:
Looks like the functionality to add, edit, and remove indicator actions has already been implemented in core/api.py;

PR #1007, opened 2018-07-12
Change bundle.get source to match other TLOs

trentprynn:
In emails/api the bundle.get() function call for the Source attempted to get data from a field called 'source_name' but in every other TLO POST the bundle attempts to get the Source from a field called 'source' and in the API Documentation one of the noted common POST parameters is 'source' not 'source_name' therefore the email API bundle call should conform to the standard and attempt to get the standard 'source' field for the Source NOT 'source_name'

### DIFF
--- a/crits/emails/api.py
+++ b/crits/emails/api.py
@@ -70,7 +70,7 @@ class EmailResource(CRITsAPIResource):
         result = None
 
         # Extract common information
-        source = bundle.data.get('source_name', None)
+        source = bundle.data.get('source', None)
         method = bundle.data.get('source_method', '')
         reference = bundle.data.get('source_reference', None)
         tlp = bundle.data.get('source_tlp', 'amber')


mgoffin:
Some quick thoughts. There is a slew of inconsistencies for the API across TLOs. For example, `sample` doesn't use the same fields email does for method or reference:

https://github.com/crits/crits/blob/master/crits/samples/api.py#L83

Getting all of these inline and identical would break the API for existing people developing against it.

I think the "right" solution would be to make a `v2` of the API that does things better. Another less good solution would be to check for current fields, but if we get nothing from them try the correct field and see if that was provided.

As for a `v2` of the API, that effort might be large enough where it's not worth doing when we could rewrite the API to be faster and better and maybe ditch tastypie completely.

brlogan:
And not just the APIs, there are a ton of inconsistencies in the TLO views and handlers as well.

A quick patch to accept either field might be good short-term, but I think the rewrite is the longer-term solution.

PR #1010, opened 2018-08-13
Attempt to chase a bug I introduced with the __ne support

ckane:
Basically, there are some cases where the value of a field in the query
string is represented as a comma-separated list of values. There is code
present earlier in this function to identify these and do the
tokenize-and-listify operation. However, that renders val as a list,
rather than string, type by the time it gets to the code to populate the
field constraint in the query dict (which is coded to call
remove_quotes(val) ).

So the solution is to test whether val is a list, and if it isn't, run
the remove_quotes(val). Otherwise, just pass the already-decoded list as
the rvalue in the assignment.

### DIFF
--- a/crits/core/api.py
+++ b/crits/core/api.py
@@ -516,10 +516,12 @@ class CRITsAPIResource(MongoEngineResource):
                                 except:
                                     val = None
                         if val or val == 0:
+                            if not isinstance(val, list):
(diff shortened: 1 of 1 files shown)


ckane:
Eric Wentz tested this fix, and was the initial reporter of the bug, merging.

PR #1012, opened 2018-08-14
Call remove_quotes only if val is a str

ckane:
In the cases where val was a str type, this would always run the
remove_quotes function. However, if the val was anything other than a
list type (such as a datetime), it would also attempt this translation.
This change limits the translation to only occur on str types, which I
think is the appropriate action.


ckane:
The issue reporter said that the problem was that any field that worked on datetime-format data was failing because remove_quotes() would attempt to run on it. After this fix, it is confirmed working.

PR #1019, opened 2018-09-26
Fix incorrect syntax for descending indexes

brlogan:
When this was last updated, it appears the wrong syntax was used to create an index with descending sort order.  Prepending a "-" to the key seems to treat it as part of the key name rather than reversing the sort order.  The confusion is understandable since I think Django's ORM and Mongoengine use the "-" in their `order_by` function, but PyMongo uses a list of tuples as the argument for its `create_index` function.


mgoffin:
Hmm, I wonder if this has any positive speed impacts as we probably sort on created frequently.

brlogan:
I think there is at least a small improvement.  Created (or "Added") is the default sort for Events and Indicators, so it should help with the initial display of those lists.

mgoffin:
One thing we might want to do is change the `-1` to `pymongo.DESCENDING`. I think they are currently the same thing, but if it ever changes in the future we'll be safe. Thoughts?

brlogan:
That would indeed be the more correct way to do it. Just need to import pymongo first.  Want me to change it?

brlogan:
Okay, I swapped `-1` for `pymongo.DESCENDING`.

PR #1020, opened 2018-09-29
Added ordering for the tastypie API

apolkosnik:
This enables ordering of the API query results.


apolkosnik:
I took the fields and dropped sources and campaigns. I think that this should do it.

apolkosnik:
I think it should also work for stable_4

PR #1021, opened 2018-10-02
Fix user objects

ManoftheSea:


ManoftheSea:
It clears up at least a few errors in:
- The interface to add threat types and attack types
- The interface for Confidence and Impact

However, I'm a little unsure of the style... Where it has `"%s" % user` to cast from the CRITsUser object to a string, sometimes it's changed to username.

mgoffin:
That was an artifact of the username string for some reason having issues being converted to a string.

As long as they’ve been changed to be the user object to be passed into a handler that can properly convert the username when it needs the string (usually for display or entering into the database) then that’s fine.

I’ll merge and if there are further issues they can get cleaned up as well.

PR #1023, opened 2018-10-12
Distinguish object from string in object->views, handlers

ManoftheSea:
definitely a mistake in objects/views passing a string to "add_new_object" when it expects a CRITsUser.


christauff:
We rolled these changes into our local CRITS code base then killed all extant .pyc, and restarted the server after the edits made above and proceeded to try to add a single Object as Indicator. 

Objects DID NOT add successfully.

Both modal windows had the same error: 
POST /objects/add
POST /objects/bulkaddinline
returned error code 500 "Internal Server Error."

Promoting an extant added Object on a Sample page from BEFORE the code changes  above gave same error: 
POST /objects/create_indicator/ -> 500 error
returned error code 500 "Internal Server Error."

And after prospective code changes were made, we can no longer add Objects in any way to a sample.

christauff:
BTW, @ManoftheSea, if you have an suggestions we could undertake this weekend, I'll have a fair bit of time to comb through recent commits and look for things that might be worth pursuing.

We felt leery of diving into wholesale handler.py, view.py and forms.py edits without some guidance, and after looking at the recent @mgoffin commits as well as yours, we see some avenues of investigation, but we've not yet grokked the user Object vs User string context in the codebase.

Regardless, thank you for your time and efforts to help us mitigate this issue. We use CRITS heavily in a production environment, and would really like to use the Master branch, instead of rolling back to Stable-4. 

:-)

christauff:
Thanks, @ManoftheSea, I very much appreciate it! I can now add Objects as Indicators from the Single Object add button! We're now able to do work again.

However, I still cannot add Bulk Objects as Indicators, and I cannot promote an existing Object to an Indicator using the plus sign in the Objects section on a Sample page.

Any ideas?

P.S. I have not joined the Slack server, but will be happy to.

PR #1032, opened 2018-11-18
username fixes

apolkosnik:
Fix update_object_data(),  toggle_favorites(), favorites()
user.username needs to be passed


mgoffin:
Isn’t this the wrong fix? We should be passing the user object and those functions should use it to get the username if they need it.

PR #1033, opened 2018-12-12
move loaded_mods out of base_context

apolkosnik:
This should lighten the load a little by moving it out of the base_context.


mgoffin:
If we are expecting a larger PR that involves base_context, we probably shouldn’t make additional changes that would introduce a merge conflict or stomp on that effort.

PR #1034, opened 2018-12-13
Form rendering improvement.

jwilczek:
Render forms when requested as opposed to rendering every form at page load via the page context processor. This update should drastically increase performance of the CRITS gui.


jwilczek:
Accidentally selected master.

PR #1040, opened 2018-12-17
Fixed edit items dialog

carterchurch:
Added add_new_action back into base_context; This must stay in base_context until the config page is modified to load forms from the dialog function.

### DIFF
--- a/crits/core/views.py
+++ b/crits/core/views.py
@@ -1296,6 +1296,12 @@ def base_context(request):
         except Exception, e:
             logger.warning("Base Context ActionsForm Error: %s" % e)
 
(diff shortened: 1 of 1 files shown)


mgoffin:
Would it be better to fix the config page instead?

mgoffin:
For sure. I suspect we can do the right long-term fix instead. I’m not positive but I suspect the amount of people on Master is a lot lower than stable and can probably hold out for the right fix.

apolkosnik:
I thought that both were affected.

carterchurch:
@mgoffin In that case, I'll close this out and look implementing this correctly

PR #1048, opened 2019-04-24
ticket_add handler fixup

pinowudi:
Shortened 'user.username' to 'user' because the function does not know how to handle the username object. It does handle the user object. This fixes patch updates to objects.


mgoffin:
I don't believe this is the correct fix. What is the problem you are seeing. If you look at the function `.add_ticket()` (found here: https://github.com/crits/crits/blob/master/crits/core/crits_mongoengine.py#L1134) it does in fact want the username string, not the user object.

PR #1049, opened 2019-05-04
Replacing len() on Querysets with .count() to make the DB do the heavy lifting 

apolkosnik:
Replacing some len() with .count(). It makes DB do work rather than Python. This also should reduce lag in those places, as Python will not pull down the results of the query, and will not have to iterate through them to count them.

Also added a note about SetOnInsert being supported by Mongoengine 8.0.0,
so that the query can be changed to run under mongoengine rather than PyMongo

### DIFF
--- a/crits/certificates/handlers.py
+++ b/crits/certificates/handlers.py
@@ -345,7 +345,7 @@ def handle_cert_file(filename, data, source_name, user=None,
     cert.reload()
 
     # run certificate triage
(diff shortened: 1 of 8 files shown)


mgoffin:
I thought MongoEngine decided the best way to perform counts using their framework was to use `len()` and to not use count? If that's not the case then this seems fine.

apolkosnik:
http://docs.mongoengine.org/guide/querying.html#counting-results

